Issue #2, opened 2025-06-04
Dataset 'mech-interp-bench/copycolors_mcqa_private_test' doesn't exist on the Hub or cannot be accessed.


c4n:
Hi,
I'm trying to play around with the MIB benchmark, I'm not sure if we are meant to access this dataset, but it does not seem to work right now.

https://github.com/atticusg/CausalAbstraction/blob/118f299d8725ce27eb7ce45b0bc4bc30daad31f0/tasks/simple_MCQA/simple_MCQA.py#L106

 datasets.exceptions.DatasetNotFoundError: Dataset 'mech-interp-bench/copycolors_mcqa_private_test' doesn't exist on the Hub or cannot be accessed.

giorgioskij:
I noticed the same thing. I sent PR #4 to add a parameter to `get_task` that fixes this.

atticusg:
Hello, thanks so much for engaging with this code base! I really appreciate it :)

I was refactoring this code base and renamed branches, which automatically closed the PR mentioned in this issue thread.

I have implemented the equivalent to this PR in the new refactored code though!

Please see the following:

https://github.com/aaronmueller/MIB/tree/main/MIB-causal-variable-track/tasks

Apologies for the poor Git etiquette! The current codebases should be stable and ready to be built upon with PRs like this one.

PR #4, opened 2025-06-04
Fixes get_task automatically fetching private dataset from HF

giorgioskij:
As mentioned in #2, `get_task(...)` tries to always fetch the private dataset from HF. This PR adds a parameter `include_private: bool = False` to get_task() to explicitly include the private dataset.


giorgioskij:
Why was this closed? do you have plans of fixing differently or to not fix at all? usually the discussion under a pr is the best place to discuss such issues!

giorgioskij:
Great! No worries, and thanks for the updates! Keep it up 🔥

PR #6, opened 2025-07-14
Update causal model visualization

AmirZur:
This PR introduces three visualization functions to `causal_model.py`:
* `display_structure` - displays model structure without inputs
* `display_forward_pass` - displays forward pass of model with base input and (optional) intervention
* `display_interchange` - displays interchange intervention between base input and counterfactual input(s)

The PR also includes a demo notebook for visualizing/understanding causal models through a toy arithmetic example.

### DIFF
--- a/causal/causal_model.py
+++ b/causal/causal_model.py
@@ -11,8 +11,20 @@ import matplotlib.pyplot as plt
 import networkx as nx
 from datasets import Dataset, load_dataset
 
+# visualization libraries
+from dash import Dash, html
+from dash.dependencies import Input, Output, State
+import dash_cytoscape as cyto
+
 from causal.counterfactual_dataset import CounterfactualDataset
 
+class DEFAULT_COLORS:
+    """default colors for displaying causal model with dash"""
+    BASE_INPUT = "#68AF9C"
+    BASE_OUTPUT = "#A0CCC0"
+    SOURCE_INPUT = "#EC3B82"
+    SOURCE_OUTPUT = "#F276A8"
+    COUNTERFACTUAL_OUTPUT = "#A59FD9"
 
 class CausalModel:
     """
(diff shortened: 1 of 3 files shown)


JadenFiotto-Kaufman:
## A golden for the held read — and why it does not catch the `_gather` regression

Pushed `3b1b283`: `tests/golden/test_cuda_graphs.py::test_a_read_off_a_replay_outlives_the_next_replay`.
The two-sided verification asked for came back **one-sided**, so the honest headline first.

### The new golden

Nothing above the unit tier kept a read past the replay that followed it. Every graph golden
compares a read inside the step that produced it. The one that holds a value across steps,
`test_inference_replay_preserves_outputs_and_forward_cache`, holds `patched_act` at step 0 —
and that value came from no graph at all: the test rolls `input_ids` every step, restaged
tokens are what `GraphExecutor._check_frozen_inputs` closes the captures for, so over the whole
test **`inference_captures: 0`, `inference_replays: 0`** (measured with a `Replay.__init__` /
`__call__` counter). Its `assert all(replay.replays >= 1 for ... in actual._inference_graphs.values())`
is an `all()` over an empty dict.

The new test holds the **intervened model's** reads, on the device, taken off one replay and
compared after the next has run under a featurizer parameter the first never saw. Three
conditions decide whether the held value is a graph's, and each is asserted rather than assumed:

* `device_reads = True` — a read copied to the host cannot alias anything.
* the *intervened* model's reads — a fit document's `original` groups are **frozen sources**
  (`_forward_group`: run once, deep-cloned), so `base_act` / `v_cf` are served from a clone and
  never from a graph. Only the `patched` group's reads come off a replay.
* fixed tokens, with the parameter moved instead — which is what a fit's own replays vary.

It asserts the graphs exist and have replayed (`1` capture, `2` replays), and that the held
reads *moved* over the later replay, so it cannot pass vacuously.

### What it is red for

Measured on `Qwen/Qwen3-8B` on one A100, it is red exactly when nothing between the graph's
static output buffers and the read copies:

| build | result |
|---|---|
| unmutated | `1 passed` |
| **`finalize_read(gather=gather_rows)` — the PR #6 state** | **`1 passed`** |
| `GraphExecutor._gather` returns the slice, no `.clone()` | `1 passed` |
| the replay's capture dict not cloned (`cuda_graphs.py:804`) | `1 passed` |
| **neither clone** | **`1 failed` — `AssertionError: read 'patched_act' was overwritten`** |
| PR #6 state *and* no capture-dict clone | `1 passed` |

So it is the golden for invariant 2 of `docs/cuda_graphs.md` — *"an inference replay's captures
and routing are cloned immediately after the replay"* — which had none. The two copies are
redundant with each other, which is why either alone still passes.

### Why it does not catch the `_gather` regression, and why I think no golden can

That same redundancy, plus the eligibility gate. `unsupported_reason` admits only
`pos: {index: -1}` reads and writes, so on **every** document the graph path accepts, the
override's `tensor[:, -1:, :].clone()` and `gather_rows(tensor, per_row)` address the same
position — and both own their storage (`gather_rows` is an advanced index). Instrumented over
the whole suite: **754 `_gather` calls, 126 of them inside a capture, zero at any position other
than the last, zero returning a value that aliases its input.** Reverting the dispatch is a
value-identical change on every graph-eligible document.

The override's reachability is therefore a unit-tier fact — pinnable only by feeding it
positions the gate forbids, which is what the unit test in `963a63c` does — and not a
golden-tier one. I have not claimed otherwise in the commit message.

### One thing worth a follow-up (not in this commit)

`test_inference_replay_preserves_outputs_and_forward_cache` is named for a replay it never
performs. Either it should stop rolling tokens (and vary the parameter, as the new test does)
so its `_inference_graphs` are real, or it should assert `actual._inference_graphs` is non-empty
so the day it stops capturing is a red and not a silent pass. Same for the second half: the two
`run_all()` calls are an eager pass and a `ForwardCache` hit, so the frozen-taps comparison
there never spans a replay either.

### Checks

* `uv run pytest -m golden -n 0 tests/golden/test_cuda_graphs.py` on `Qwen/Qwen3-8B`: `24 passed in 104.06s`.
* CPU gate unaffected — the new test is `golden`-marked, and `tests/golden -m "not golden" --collect-only`
  collects nothing from the module (24 deselected). `tests/neural/engines/nnterp_engine` `716 passed`,
  `tests/neural/engines/pytorch_hooks` `1593 passed, 4 skipped`.

PR #7, opened 2025-07-16
RevNet Integration

densutter:
This is a suggested approach for integrating RevNets into the existing DAS implementation on MIB.


can-goodfire:
Thanks for contributing @densutter ! Causalab has moved a lot since, so closing this PR for now. Still happy to collaborate!

PR #15, opened 2026-02-01
Fix badge link in README.md

ScottBrenner:
Current badge doesn't render properly, correcting with status badge from https://github.com/goodfire-ai/causalab/actions/workflows/test.yml

Reference https://docs.github.com/en/actions/how-tos/monitor-workflows/add-a-status-badge#using-the-ui


maxsloef-goodfire (approved):
nice catch, ty!

PR #21, opened 2026-08-20
Metric answers: a token_form knob, and featurizers that reach the run's device

can-goodfire:
Two bugs on the headline path, found running the causal-analysis pipeline against this branch. Neither is reachable by the corpus tests, which run tiny-random on CPU.

Stacked on #20 — targets `can/protocol-refactor`, not `main`.

---

## Bug 1 — a metric silently scored a token the model never emits

`column_token_id` returns the first single-token candidate, trying the space-prefixed form first. Under gpt2, both forms of a punctuation answer are single tokens and they differ:

| answer | `" " + s` | `s` |
|---|---|---|
| `?` | 5633 | **30** |
| `.` | 764 | **13** |

So it scored 5633 while the model emitted 30. A `match`-based IIA read **0.000 at all 48 layers** of a real gpt2-xl scan — no error raised, which a pipeline gate reads as a dead stage rather than a failure. There was no escape hatch: every string-answer kind (`logit_diff`, `token_logit`, `cross_entropy`, `match`, `class_probs`) resolves through it, and `train.metric_tensor` too. Only `kl` and `top_k` avoid it, and neither takes a string answer.

**The space-prefixed preference is correct** for the common case — `" Monday"`, IOI names, MCQA letters — and `tests/tasks/IOI/test_readout_token.py` pins it: 53 of 99 IOI names break under the opposite order. The bug is that a caller had no way to say otherwise.

**Fix — an optional per-metric `token_form`:** `auto` (default) | `bare` | `space_prefixed`.

```json
"iia": {"kind": "match", "of": "logits", "expected": "answer", "token_form": "bare"}
```

- **Per-metric scalar, not a literal token id** — these fields name *dataset columns* (`expected: "answer"`), resolved per row; an integer would mean one constant token for every row and would collide with the column-reference check in `loader.py`.
- **Not per-column** — a metric's columns come from one output space and share a convention. The exception, a `class_probs` mixing spaced and bare members, stays on `auto`; documented.
- **Not in `MetricSpec.fields`** — `loader.py` treats every string there as a column name and would reject it. Refused on `kl`/`top_k` (`P3`), and not sweepable (`V14`).

**Back-compat is exact.** `auto` is today's resolver and is the default, so every existing document is unchanged and `corpus_digests.json` is untouched — canonicalization deliberately does not materialize the default, so an omitted key and an authored `"auto"` stay byte-identical.

**Not raising on ambiguity:** gpt2 makes `" Monday"`/`"Monday"` and half the IOI names both single-token, so refusing would break the common case. Instead `column_token_ids` warns **once per column**, naming up to three ambiguous answers with both ids.

## Bug 2 — trainable featurizers never reached the run's device

`device` appeared **zero** times in `featurizers.py`, so:

```
causalab run <das_doc>.json --device cuda
RuntimeError: ... mat2 is on cpu, different from other tensors on cuda:0   # Subspace.featurize
```

Every document with a featurizer — DAS, DBM, das-apply, PCA, gates, SAE — was CPU-only. That is the headline analysis node.

**Fix at the root:** `build_stack` takes `device` (default `"cpu"`, so CPU callers are unchanged) and applies `stage.to(device)`. `nn.Module.to` carries parameters *and* buffers, so the loaded-from-file stages come along; the two `PointExecutor` sites pass `self.bundle.device`.

- **Construction stays on CPU** so the seeded init is bit-identical across devices. There is a test for it.
- **Dtype left alone** — every stage already casts at the boundary, so featurizers stay fp32 against a bf16 backbone. Forcing bundle dtype would change numerics on every non-fp32 run.

---

## Verification

On a real H100, prompts pre-probed so only genuine bare-`?` answers were scored:

| `token_form` | mean `match` |
|---|---|
| `auto` | 0.000 |
| `space_prefixed` | 0.000 |
| **`bare`** | **1.000** |

Corpus 04 (DAS) `--device cuda` → rc 0 with a fitted rotation; reverting the two files reproduces the crash verbatim.

**Tests:** new `test_featurizer_device.py` (11 cases, 20 with a card) parametrized over available devices — reverting only `stage.to(device)` turns 8 red. `test_metrics.py` +9 on the real gpt2 tokenizer (tiny-random's 1000-token vocab cannot express the bug). `test_schema.py` +6. `pytest -m "not golden"` → **1447 passed**.

**CI `lint` is red and inherited** — 5 × `E741` in `tests/golden/test_paper_goldens.py`, untouched here; the base branch fails identically and ruff cannot auto-fix E741. `test` is green.

## Found, not fixed

`Subspace.__init__` takes a `seed` that `_build_stage` never passes, so every subspace init uses seed 0 regardless of `train.seed` — a sweep on the seed may share one initial rotation. Fixed separately on `can/subspace-seed-threading`.


can-goodfire:
## GPU verification — real H100 (Reno, SLURM jobs 1380792 / 1380910)

`NVIDIA H100 80GB HBM3`, torch `2.9.0+cu128`, branch cloned fresh at `553899b`.

### A. Device unit tests, `cuda` parametrization live — **20 passed**

Every `[*-cuda]` case ran (not skipped): all five stage kinds, the composed chain, the loaded-from-file kinds, the seeded-init equality, and the `featurize`/`inverse` round trip.

### B. Corpus 04 (DAS — trained subspace featurizer) with `--device cuda` — **rc 0**

```
saved ce.parquet / iia.parquet / rot.safetensors
fitted rot.weight: (16, 4) torch.float32
```

The headline analysis node now trains and saves on GPU.

### C. Same command, `featurizers.py` + `executor.py` reverted to `HEAD~1` — **reproduces the reported crash**

```
File ".../causalab/neural/pytorch_hooks/featurizers.py", line 83, in featurize
    f = x.to(q.dtype) @ q
RuntimeError: Expected all tensors to be on the same device, but got mat2 is on cpu,
              different from other tensors on cuda:0 (when checking argument in method wrapper_CUDA_mm)
```

Byte-for-byte the error in the report, so B is a fix and not an accident of the environment.

### D. Punctuation `match` on real gpt2, `--device cuda`

Prompts were **probed first** and only those whose greedy next token really is the bare `?` (id 30) were kept — 8 of 16 — so the metric has a right answer to find. (Instructive miss: gpt2 merges `?"` into a single token 1701, so a *quoted* question never exposes the bare row at all.)

```
match  auto (today's resolver)        -> mean 0.000  [0,0,0,0,0,0,0,0]
match  token_form='space_prefixed'    -> mean 0.000  [0,0,0,0,0,0,0,0]
match  token_form='bare' (the fix)    -> mean 1.000  [1,1,1,1,1,1,1,1]
```

The identically-0.000 read from the original gpt2-xl scan, reproduced on demand and then fixed by one declarative key. The warning fired once for the column, as designed:

```
UserWarning: metric match.expected: 1 of 1 distinct answers are ambiguous under this
tokenizer — both forms are single tokens and they name different rows
('?' → 5633 (space-prefixed) vs 30 (bare)). token_form='auto' took the space-prefixed
form; set the metric's token_form to 'bare' or 'space_prefixed' to say which one the
model actually emits.
```

can-goodfire:
**Note on the red `lint` check — it is inherited, not introduced here.**

The failure is 5 × `E741 Ambiguous variable name: \`l\`` in `tests/golden/test_paper_goldens.py` (lines 201, 217, 240, 261, 268), a file this PR does not touch. The base branch fails the same way: [run 32383692683](https://github.com/goodfire-ai/causalab/actions/runs/32383692683) on `can/protocol-refactor` @ `c022c19` is `lint failure` / `test success`, and the auto-fix commit that followed (`0d78aae`) could not fix E741, which ruff does not auto-fix.

Everything this PR adds or edits is clean under both `ruff check` and `ruff format --check`. `test` and `Test Results` are green.

The one-line fix (`for l in` → `for layer in`, or a `per-file-ignores` entry for the golden tests, matching the existing `"*.ipynb" = ["E402", "E741"]` rule) belongs with whoever owns the paper-goldens tier, not in this bugfix.

PR #30, opened 2026-08-20
Dataset-serialization seam: task-generated tables, answer-form groups, per-row position columns

### LINKED ISSUE #23: Dataset-serialization seam: resolver contract, task-generated tables, and the two task-supplied columns (PR #20 Q8/Q9/Q13)
## Why

Protocol documents name datasets by ref (`"dataset": "weekdays/train"`), but nothing in the repo produces a real table. Every ref today resolves to a hand-authored JSON fixture (`tests/protocol/fixtures/data/`, and `tests/golden/fixtures/data/` written by bespoke `tests/golden/fixtures/generators/gen_*.py`), while `causalab/tasks/` holds nine task packages with causal models and counterfactual generators that no document can reach. PR #20 raised this as spec question **Q9 (dataset resolution)**, plus two questions that are really *consumers* of the same seam — **Q8** (`match` semantics when an answer has synonyms or is multi-token) and **Q13** (per-row / dynamic positions).

What is blocked on it:

- **The task-driven end-to-end IIA pin** — the only end-to-end IIA number the stack ever had (task config → positions → edit → plan → one IIA number), retired in the refactor as `test_walking_skeleton.py` (`docs/test_migration.md`, that row and the "Task-driven end-to-end IIA pins" gap).
- **Every real-task preset/workflow** — `causalab/configs/workflows/weekdays_8b.json` and `causalab/configs/methods/weekdays_*.json` name `weekdays/*` refs that exist only inside the test fixture root.
- **Recomputing the `subject_object_relations` curation table.** Its producer (`data/curation_sweep.py`) was deleted with the runner/methods stack while its numbers stay load-bearing — they select `config.py`'s default relation and the relation each pinned tier uses (see that task's README, "Curation"). Re-expressing it as a document campaign needs this seam *and* a first-token/prefix match form (Q8).
- **MCQA's dynamic positions** — `causalab/tasks/MCQA/token_positions.py:51`, `correct_symbol` / `correct_symbol_period`, are inexpressible in v1.

## What already exists (so this is smaller than it looks)

- `causalab/protocol/resolve.py:62` — the `DatasetResolver` protocol (`digest`, `columns`) and `FileDatasets` (`resolve.py:280`), JSON tables under a data root, sha256 of the file bytes as the content digest.
- `causalab/protocol/canonical.py:177` `_canon_data` already stamps the dataset digest into the canonical form (§2.2/§7), and `causalab/protocol/loader.py:267` `check_data_columns` already checks metric/field column references under `validate --data`.
- `causalab/neural/pytorch_hooks/encoding.py:113` `variable_value` already resolves prompt variables from a `<column>_variables` sibling *or* a same-named plain column — Q9's proposed convention, implemented.
- `causalab/causal/causal_model.py` already owns the task-side vocabulary the two column consumers need: `build_output_tokens` (`{value: [forms]}`), `MATCH_MODES = ("exact", "prefix")`, `derive_checker`, and `label_counterfactual_data` (runs the interchange through the causal model to produce the post-intervention label). `causalab/causal/causal_utils.py:711` `form_groups` already dedups form groups.
- `causalab/protocol/canonical.py:387` `_canon_train` is the precedent for materializing a defaulted field into the canonical form (optimizer defaults, precision, seed) — reused below.

## Design

Guiding principle (from Q8/Q13): **anything per-row or task-semantic is computed by the task at dataset-build time and serialized as a column.** Documents reference columns; they never compute. The protocol layer stays torch-free and stdlib-only, and load-time digests never depend on importing task code or a tokenizer.

### 1. Close the resolver contract — `causalab/protocol/resolve.py`

`rows(ref) -> list[dict]` is already *required* at run time but only duck-typed: `causalab/neural/pytorch_hooks/backend.py:211` does `getattr(datasets, "rows", None)` and raises if absent. Add `rows` to the `DatasetResolver` protocol so the contract is what execution actually needs (`digest` / `columns` / `rows`), and drop the getattr probe.

No new resolver class. Refs stay local paths under `--data-root`, resolved by `FileDatasets`; task tables get there by being **built and written**, not generated at load (see 2). Rationale: `resolve.py` is deliberately stdlib-only and the pure verbs (`validate`/`explain`/`digest`) are torch-free — a resolver that imported `causalab/tasks/` at load would make a document's digest depend on task code and a tokenizer being importable.

Spec text fix while here: §2.2 says a ref is a "local path or HF key". Only local paths are implemented, and resolving a Hub ref at load would put a network fetch inside digest computation. Narrow the sentence to local paths and say Hub datasets enter by being materialized first.

### 2. Task → serialized table builder — `causalab/tasks/serialize.py` + `scripts/build_task_dataset.py`

A library function plus a thin script (same shape as `scripts/update_task_pins.py` and the `tests/golden/fixtures/generators/gen_*.py` family, which this generalizes):

```
serialize_counterfactual_dataset(task_name, task_cfg, *, n, seed, target_variables,
                                 generator="generate_dataset") -> list[dict]
```

It calls `load_task` → the task's `counterfactuals.<generator>(causal_model, n, seed)` → `causal_model.label_counterfactual_data(examples, target_variables)`, and emits one row per example with the column vocabulary the existing fixtures already use, so no document has to learn new names:

| column | from |
|---|---|
| `input` | base trace's `raw_input` (the rendered prompt) |
| `counterfactual_inputs` | list of counterfactual traces' `raw_input` |
| `base_answer` | base trace's `raw_output` |
| `cf_answer` | counterfactual trace's own `raw_output` |
| `label` | `label_counterfactual_data`'s interchange output — the expected answer *after* the intervention |
| `input_variables`, `counterfactual_inputs_variables` | the traces' variable settings, in the `<column>_variables` convention `variable_value` already reads (the counterfactual sibling is a list, indexed by `counterfactual_inputs[j]`) |
| `<answer>_forms` | `output_tokens[target_variable]` forms for that row's value, when the causal model declares them (Q8, see 3) |
| task-declared position columns | an optional `POSITION_COLUMNS` task export (Q13, see 4) |

Determinism is the point: generators already snapshot/restore the global RNG, and the writer uses `json.dumps(rows, indent=1, sort_keys=True)`, so the same (task, cfg, n, seed) produces byte-identical bytes and therefore the same content digest on any machine.

A sidecar `<ref>.manifest.json` records `task`, `task_cfg`, `n`, `seed`, `generator`, `target_variables`, the causalab commit, and the table digest. It is provenance only — never read at load, never digested — and is the machine-readable half of what `causal_model.key` records as a name (§2.1).

**Correctness note worth putting in the docs:** IIA should score against `label`, not `cf_answer`. They coincide only when the interchange replaces every variable the answer depends on — which is why `tests/protocols/02_interchange_im.json`'s `expected: "cf_answer"` looks right today. Emitting both columns and documenting the difference closes a real trap.

### 3. Answer-form groups and a first-token `match` (Q8)

- `match`'s `expected` field may name a **list-valued** column: the row's forms are alternatives, and the metric scores 1.0 when the argmax token is any of them. A string-valued column keeps today's exact behaviour, so nothing existing changes.
- `match` gains an **optional** `mode` field, `"exact"` (default) or `"first_token"`. `first_token` resolves each form to its *first* token id instead of requiring the form to be single-token — which is what "prefix" matching means when you only have logits at one position, and it is precisely the form the `subject_object_relations` answer space needs (multi-token objects such as `Washington D.C.`). It mirrors the task-side `MATCH_MODES = ("exact", "prefix")` and reuses `column_token_id`'s space-prefixed-first candidate order (`causalab/neural/pytorch_hooks/metrics.py:26`).
- `mode` must be **opt-in per document**: `first_token` credits a prefix, so on an answer space that is not first-token-distinct it over-credits. Whether a table's answer space is first-token-distinct is a task-data property (the `subject_object_relations` README already reports it per relation) — not something the metric can check.
- Case folding stays out of the metric. Forms are compared as token ids, so folding at scoring time would mean decode-and-compare; a task that wants case-insensitivity serializes the case variants as forms. This keeps "reproducing historical numbers" a task-data concern, as Q8 intended.
- Mechanics: `METRIC_FIELDS` entries are currently all mandatory (`causalab/protocol/schema.py:1130`). Add an `OPTIONAL_METRIC_FIELDS` table alongside it, and materialize the default into the canonical form the way `_canon_train` materializes optimizer defaults, so an omitted `mode` and an explicit `"exact"` digest identically and no digest depends on an unstated default.

### 4. Per-row position columns (Q13)

Positions gain a `column` spelling wherever `variable` may appear — as the anchor (`{"column": "correct_symbol"}`) and inside `scope` / `relative_to` (`{"index": 1, "relative_to": {"column": "correct_symbol"}}`, which is exactly MCQA's `correct_symbol_period`). The row's value is a string and resolves through the existing unique-occurrence token run, per row, like a prompt variable.

Why a new spelling when `variable_value`'s plain-column fallback nearly does this already: `column` is explicit (only the row's top-level columns, never the `<column>_variables` sibling), so `validate --data` can check the reference the way it checks metric columns — today a position that leans on the fallback fails at run time, deep in the backend, with no load-time check. It also keeps tables tokenizer-independent, which matters because the same table is meant to run under different models.

Integer index columns (a task computing token indices with a tokenizer) are deliberately **not** in v1: they would bind a table to one tokenizer, and the string form covers both MCQA cases. If a later case needs them, `{"column": …}` can accept an int payload without a new spelling.

## Implementation plan

1. **Contract** — `rows` into `DatasetResolver`; delete the `getattr` probe in `pytorch_hooks/backend.py`. Narrow §2.2's "or HF key". *Test:* an env whose resolver lacks `rows` fails to type-check/raises at the seam, not mid-run.
2. **Builder** — `causalab/tasks/serialize.py` + `scripts/build_task_dataset.py` (`--task`, `--set k=v` for the config, `--n`, `--seed`, `--target-variable`, `--generator`, `--out`). *Tests:* round-trip determinism (two builds are byte-identical); the committed fixture below is byte-reproducible by the script (the regeneration guard `update_corpus_digests.py` and `gen_*.py` already use); columns match the documented vocabulary; `label` differs from `cf_answer` on a partial-interchange case if a task offers one.
3. **`match` groups + `first_token`** — schema (optional fields, canonical default), `pytorch_hooks/metrics.py`, spec §2.10 table. *Tests:* in `tests/neural/pytorch_hooks/test_metrics.py`, a group column scores 1.0 for a non-argmax-but-listed synonym and 0.0 otherwise; `first_token` credits a multi-token form whose first token wins and `exact` refuses that same form (today's error stays the default behaviour); a validation test that an unknown `mode` is a load error.
4. **`column` positions** — `_parse_position_spec` / `_parse_var_ref` in `schema.py`, resolution in `pytorch_hooks/encoding.py`, the reference check in `loader.py::check_data_columns`, spec §2.3 table. *Tests:* a `column`-anchored read and a `relative_to: {column: …}` read resolve per row against a fixture with differing values per row; a missing column is a `validate --data` error with a rule number.
5. **The end-to-end IIA pin** — build the weekdays table (`natural_domains_arithmetic`, `domain_type=weekdays`, `n=4`, `seed=0`, target `result`) as a **new committed ref** under `tests/protocol/fixtures/data/`; add a corpus-02-shaped interchange document over it and a test — `tests/neural/pytorch_hooks/test_end_to_end_iia.py`, named for what it pins rather than reviving "walking skeleton" (Cockburn's term for a thin end-to-end slice; it was milestone WS1 of the superseded nnterp rebase plan) — that runs it on `tiny-random` through the CLI at two layers, pinning IIA plus a non-zero `logit_diff` (the retired test's own guard against a silently-inert interchange). Do **not** regenerate the existing hand-written fixtures — that would churn `tests/protocol/corpus_digests.json` and bury the real diff.
6. **The `subject_object_relations` case, load-only** — a test that builds tables for a few relations into `tmp_path`, then loads and validates a baseline document sweeping `data.base.dataset` over them with a `first_token` `match` (`dataset` is already sweepable — `_wrapped(..., allow_sweep=True)`). CPU, no model: it proves the seam covers the curation shape without pulling a GPU campaign into this issue.
7. **Docs** — §2.2 (the `rows` contract, generated-table provenance, `label` vs `cf_answer`), §2.3 (`column`), §2.10 (`match` groups + `mode`); `causalab/tasks/README.md` (the serialization convention and the `POSITION_COLUMNS` export); `docs/test_migration.md` (the retired-test rows and the gap entry — the gap is now called "task-driven end-to-end IIA pins"; "walking skeleton" survives only where it names the deleted files); `docs/CODEBASE.md` (the builder's home); the `subject_object_relations` README (what still stands between the table and a recompute).

## Acceptance

- [ ] `DatasetResolver` declares `digest` / `columns` / `rows`; the backend no longer probes for `rows` with `getattr`.
- [ ] `scripts/build_task_dataset.py` turns a task + counterfactual generator + (n, seed) into a serialized table with a provenance manifest, byte-identically on repeat runs.
- [ ] A `match` over a synonym-group column and a `first_token` `match` over a multi-token form each load, validate, and score correctly on fixture data; an omitted `mode` and an explicit `"exact"` produce the same digest.
- [ ] A document whose read positions come from a `column` (anchor and `relative_to`) loads, validates, and executes; a missing column is a load-time `validate --data` error.
- [ ] The task-driven end-to-end IIA pin is restored as `tests/neural/pytorch_hooks/test_end_to_end_iia.py`: a built weekdays table drives an interchange document end-to-end on `tiny-random` through the CLI, with pinned IIA values and a non-inert patch assertion.
- [ ] A swept baseline document over several `subject_object_relations` relation tables loads and validates on CPU.

Honesty note on the pin: the retired pins were produced by a different scoring path (`task.checker` over generated strings, and for the GPU half a chat template that v1 cannot express). Expect the restored numbers to be **captured fresh**, with the old tiny-random values used as a cross-check rather than a target; the coherent-model half belongs to the golden tier, not here.

## Out of scope

- **A lazily-generating resolver** — rejected above (load-time digests must not depend on importing task code).
- **Hub-hosted dataset refs** and **parquet tables** — the format seam can grow later; nothing in flight needs it, and JSON keeps digests trivially reproducible.
- **Integer position columns** — deferred with a reason (see 4).
- **Running the `subject_object_relations` curation sweep** and re-pinning any coherent-model number — that is a GPU campaign, and belongs with the preset/golden work, not with the seam.
- **Chat-template documents** — an acknowledged v1 gap, unrelated to this seam.

## Decisions worth confirming before I start

1. **Builder, not resolver** (design 1/2): task tables are materialized by a script and read by `FileDatasets`. This is the one structural choice here; everything else follows from it.
2. **`match` grows an optional `mode` field** rather than a new `first_token_match` kind — smaller vocabulary change, and it introduces the optional-metric-field mechanism that later kinds will want anyway.
3. **`label` as the IIA column** in new documents, with `cf_answer` kept for the counterfactual prompt's own answer.

can-goodfire:
Closes #23. **Stacked on `can/protocol-refactor` (#20)** — review that first; this PR's diff is only the commits above its tip.

## What this is

Protocol documents named datasets by ref, but nothing in the repo produced a real table: every ref resolved to hand-authored fixture JSON, while `causalab/tasks/` held nine task packages no document could reach. This closes that seam and the two column channels that depend on it (PR #20 spec questions **Q9**, **Q8**, **Q13**), then uses them to restore the end-to-end IIA anchor the refactor retired.

Six commits, each reviewable on its own:

| commit | what |
|---|---|
| `docs:` name the retired pin | vocabulary fix requested in review: "walking skeleton" → "task-driven end-to-end IIA pin" |
| `protocol:` rows() joins the contract | `DatasetResolver` = digest / columns / rows |
| `tasks:` serialize a table | `causalab/tasks/serialize.py` + `scripts/build_task_dataset.py` |
| `protocol:` groups, first_token, columns | Q8 + Q13, and one digest migration |
| `tests:` restore the IIA pin | corpus 10 + `test_end_to_end_iia.py` + the curation-shape test |
| `style:` / `docs:` | a drive-by lint fix and the docs pass |

## The design decision worth arguing about

**A build step, not a generating resolver.** Task tables are *materialized* by a script and read back by the existing `FileDatasets`. The alternative — a resolver that drives `causalab/tasks/` at load — would make a document's digest depend on importing task code and a tokenizer, and `protocol/resolve.py` is deliberately stdlib-only so that `validate` / `explain` / `digest` stay torch-free. Writing the table first also keeps one table runnable under different models, and makes the digest reproducible from recorded parameters rather than from an import graph.

So: `serialize_counterfactual_dataset(...)` → deterministic bytes → `<ref>.json` plus a `<ref>.manifest.json` provenance sidecar that resolution ignores. `--check` rebuilds and fails instead of writing, which is how a committed table stays honest.

## What the row vocabulary decides for you

Every spec question here landed on the same principle, now written into §2.2: **anything per-row or task-semantic is computed when the table is built and serialized as a column.** Documents reference columns; they never compute.

One consequence is worth flagging as a **bug class this PR closes**: `label` — the answer *after* the interchange, from `CausalModel.label_counterfactual_data` — is now its own column next to `cf_answer`, the counterfactual prompt's own answer. They coincide only when the intervention replaces every variable the answer depends on, which is why corpus 02's `expected: "cf_answer"` looks right and is right *there*, and would be quietly wrong for a partial interchange. `tests/tasks/test_serialize.py` pins a case where they differ (intervene on `number` alone).

## Q8 — answer forms, and grading a multi-token answer

`match`'s `expected` may name a **list-valued** column (the forms are alternatives; the argmax matching any of them scores), and gains an optional `"mode": "first_token"` that credits a form's first *content* token instead of demanding the form be one token.

Two things I want on the record:

- **The sentencepiece trap.** `" Thursday"` encodes as `▁ + Th + urs + day`. Crediting the *first* id would credit the bare-space piece, which every space-prefixed answer shares — so every answer would match everything. `column_first_token_id` skips whitespace-only pieces, which lands on the token the model actually emits in context (verified: `"A: Thursday"` → `… Th urs day`). There's a test for the trap, not just for the happy path.
- **`first_token` is opt-in and always will be.** It over-credits an answer space that is not first-token-distinct, and whether a table's answers *are* distinct is a property of the dataset, invisible to the metric. `test_end_to_end_iia.py` asserts distinctness for the table it grades rather than assuming it.

Case folding stays out of the metric: the comparison is between token ids, so folding here would mean decode-and-compare. A task that wants case-insensitivity serializes the casings as forms — which is what "reproducing historical numbers becomes a task-data concern" meant.

### Digest migration (§7)

`mode` is the first *optional* metric field, materialized into the canonical form the way `train.optimizer` defaults are, so an omitted `mode` and an authored `"exact"` are one document. That moves **five pinned digests**: corpus `02`/`07`, and the golden documents `drift_interchange`, `drift_locate_scan`, `hours_baseline`. Both pin files were regenerated with their own scripts (`update_corpus_digests.py`, `update_golden_digests.py`) in the tests commit — a loader migration, deliberately, not a silent re-pin. Note the alternative was for two spellings of the same document to digest differently, which seemed worse.

## Q13 — per-row positions

A position may resolve per row from a top-level **column**, as the anchor (`{"column": "c"}`) or inside `scope` / `relative_to` — i.e. exactly MCQA's dynamic `correct_symbol` and the token after it.

The honest framing: the mechanism nearly existed, since prompt variables already fall back to a same-named column. What `column` adds is (a) an explicit reference `validate --data` can check, instead of one that fails deep in the backend mid-run, and (b) role-independence — a column is a property of the row, so it reads the same from either side of a pair. And **no new task export was needed**: the values Q13 wanted are already causal-model variables (MCQA computes `answer`), so serializing the trace's variables covers it. A task needing a genuinely new derived column adds it as a causal-model variable, where it belongs.

Integer token indices stay out of v1 deliberately — they would bind a table to one tokenizer, and a task that can compute an index can serialize the substring instead. Rules 8/9 treat column positions conservatively: two *different* columns can hold the same string on one row, so unlike two template variables nothing is provable at load.

## The restored anchor, and what its numbers are worth

`tests/neural/pytorch_hooks/test_end_to_end_iia.py` drives corpus document `10_task_table_iia_im.json` over the generated weekdays table through the real CLI, at the retired test's own n=4 / seed=0.

**Values are captured fresh, not carried over** — the old test scored generated *strings* through `task.checker`, a document scores an argmax against declared forms, and the retired coherent-model half used a chat template v1 cannot express. Flat IIA on tiny-random is the same shape the retired pins had (a randomly-initialized model answers nothing), so the pins that actually discriminate are:

- the argmax token at the answer position,
- the largest logit change the swap causes — the retired test's own `logits_delta` guard, kept for the same reason: a silently-inert interchange must not pass by matching zeros,
- the subject read's per-row widths `[1, 1, 3, 3]`, which only per-row column resolution produces (`Sunday`/`Monday` are one sentencepiece piece, `Thursday` is three).

I mutation-checked the last one: pinning `column_value` to a constant makes the run *refuse* (`'Sunday' occurs 0 times in 'Q: What day is three days after Monday?'`), so the pin is not decorative.

## Not in this PR, on purpose

- **Running the `subject_object_relations` curation sweep.** The seam plus `first_token` makes it expressible, and `tests/tasks/subject_object_relations/test_curation_documents.py` pins the campaign's shape on CPU (tables per relation, one document sweeping `data.base.dataset`, every column reference checked). Re-measuring the numbers is a coherent-model GPU run and belongs with that tier; the README now says so instead of saying "not recomputable".
- **Hub-hosted refs and parquet tables** — the format seam can grow when something needs it; JSON keeps digests trivially reproducible.
- **Integer position columns** — see above.

## Two drive-bys, called out so they're not a surprise

1. `tests/golden/test_paper_goldens.py` had five `E741`s that ruff cannot autofix, so `pre-commit run --all-files` (CI's lint job) failed on the branch independently of this change. Renamed `l` → `i` in three comprehensions; its own commit.
2. `docs/CODEBASE.md` listed a `pytorch_hooks/positions.py` that does not exist — position resolution lives in `encoding.py`. Fixed in the docs commit.

## Verification

- `uv run pytest -m "not golden"` → **1450 passed**, 8 deselected.
- `uv run pre-commit run --all-files` → clean.
- Not run: the `golden` tier (GPU). Nothing here touches a golden *value*; the three golden documents whose digests moved are pinned by `tests/golden/test_structural.py`, which runs on CPU and passes.


can-goodfire:
@claude This is a credential smoke test for the broken `claude-review` workflow (it fails on the first API call with `is_error: true` and no message, on every PR). Please reply with just `OK` and the model you are running as — do **not** review the diff.

can-goodfire:
About the red `claude-review` check here: it is **not this PR**. That workflow has failed on every non-draft PR in the visible history (`can/tensor-bundle-entries`, `can/protocol-all-positions`, `can/protocol-refactor`), always `num_turns: 1`, ~2s, `is_error: true`, with no API error in the log. I reproduced the same signature from a second entry point — the `@claude` comment above ran `claude.yml` from `main` with the same token ([run](https://github.com/goodfire-ai/causalab/actions/runs/32469715455)) — which rules out anything specific to the review workflow and leaves the credential and the shared `--model opus` pin.

Fix proposed separately in #31 (against `main`, where the workflow lives): drop the model pin, and print the execution log on failure so the actual API error is visible. I briefly pushed that change here and reverted it — the action *self-skips as a passing check* when a PR modifies its own workflow file, so carrying it in this PR would have turned the check green without testing anything.

`test` and `lint` on this PR are green.

PR #31, opened 2026-08-21
ci: unpin the model on both Claude workflows, and print why a run failed

can-goodfire:
Both Claude workflows in this repo are broken, and have been for as long as the run history shows. This unpins the model they share and makes the next failure readable.

## The symptom

`claude-code-review` has failed on **every non-draft PR** in the visible history — `can/dataset-seam`, `can/tensor-bundle-entries`, `can/protocol-all-positions`, `can/protocol-refactor` (2026-08-18 onward; everything earlier is `skipped`, i.e. draft PRs). Always the same shape:

```json
{"type": "result", "subtype": "success", "is_error": true,
 "duration_ms": 1960, "num_turns": 1, "total_cost_usd": 0}
```

One turn, ~2 s, no API error anywhere in the log. The tracking comment says only *"Claude encountered an error after 2s"*.

## Isolating it

I reproduced it from a second entry point: an `@claude` mention on a PR, which runs `claude.yml` **from `main`** — same token, different workflow, different prompt, no `track_progress`, no inline-comment tool. Identical signature (`num_turns: 1`, 2059 ms, `is_error: true`, [run](https://github.com/goodfire-ai/causalab/actions/runs/32469715455)).

Two different entry points failing identically rules out anything specific to the review workflow — its prompt, `track_progress`, the inline-comment MCP tool. What the two share is **the credential** and **`--model opus`**.

## The change

1. **Drop the `--model opus` pin** from both workflows. The alias resolves locally — the init line reports `"model": "claude-opus-5"` — and then fails on the first API request if the credential is not entitled to that model, which in this log is indistinguishable from a bad token. Without the pin, the action's default follows what the credential can actually use.
2. **Print `claude-execution-output.json` on failure.** The action declares an `execution_file` output and nothing read it, which is why the actual API error never reached the log. This is the part I'd keep regardless of what the cause turns out to be.

## Two caveats worth reading before merging

- **This PR's own checks prove nothing.** The action refuses to run when a PR's workflow file differs from the default branch's copy, and **self-skips as a passing check** ("Action skipped due to workflow validation error… your workflow will begin working once you merge your PR"). So `claude-review` will go green on this PR without executing. That is also why I could not test the fix before proposing it — a workflow change can only take effect on `main`.
- **If it still fails after merge, the remaining suspect is the token.** `CLAUDE_CODE_OAUTH_TOKEN` was last set 2026-06-11; rotating it means `claude setup-token` and updating the secret, which only its owner can do. The diagnostic step added here will name the cause instead of leaving it a guess.

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -46,5 +46,25 @@ jobs:
 
             After conducting this review, read CLAUDE.md and conduct a second review pass with those additional considerations in mind.
 
+          # No --model: the action's default follows what the credential is
+          # entitled to. A pinned alias resolves locally and then fails on the
+          # first API call if the token cannot use that model — which is what
+          # this workflow did on every PR, indistinguishably from a bad token.
           claude_args: |
-            --allowedTools "mcp__github_inline_comment__create_inline_comment,Bash(gh pr comment:*),Bash(gh pr diff:*),Bash(gh pr view:*)" --model opus
+            --allowedTools "mcp__github_inline_comment__create_inline_comment,Bash(gh pr comment:*),Bash(gh pr diff:*),Bash(gh pr view:*)"
+
+      # The action reports `is_error: true` without the API error that caused
+      # it, so a failure here is otherwise unreadable. The execution log has
+      # the actual message.
+      - name: Surface the execution log on failure
+        if: failure()
+        run: |
+          log="${{ steps.claude-review.outputs.execution_file }}"
+          [ -f "$log" ] || log="${RUNNER_TEMP}/claude-execution-output.json"
+          if [ -f "$log" ]; then
+            echo "::group::claude-execution-output.json"
+            cat "$log"
+            echo "::endgroup::"
+          else
+            echo "no execution log at $log"
+          fi
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -47,5 +47,21 @@ jobs:
           # See https://github.com/anthropics/claude-code-action/blob/main/docs/usage.md
           # or https://docs.claude.com/en/docs/claude-code/cli-reference for available options
           # claude_args: '--allowed-tools Bash(gh pr:*)'
-          claude_args: '--model opus'
+          # No --model here either: see claude-code-review.yml.
+
+      # The action reports `is_error: true` without the API error that caused
+      # it, so a failure here is otherwise unreadable. The execution log has
+      # the actual message.
+      - name: Surface the execution log on failure
+        if: failure()
+        run: |
+          log="${{ steps.claude.outputs.execution_file }}"
+          [ -f "$log" ] || log="${RUNNER_TEMP}/claude-execution-output.json"
+          if [ -f "$log" ]; then
+            echo "::group::claude-execution-output.json"
+            cat "$log"
+            echo "::endgroup::"
+          else
+            echo "no execution log at $log"
+          fi
 


can-goodfire:
Two data points from a parallel dig at the same failure, both of which tighten this PR rather than change it.

**1. The one `success` in the review history was also a self-skip — so the real count of executed reviews is zero, not "a few.”** `Claude Code Review` run [32469550141](https://github.com/goodfire-ai/causalab/actions/runs/32469550141) (today 09:47, `can/dataset-seam`) is the only `success` the run history holds. Its log:

```
##[warning]Skipping action due to workflow validation: Workflow validation failed.
The workflow file must exist and have identical content to the version on the
repository's default branch.
...
Action skipped due to workflow validation error.
```

That is 6a08d83 (this fix, committed on `can/dataset-seam` before being reverted in 4079b56) taking effect — and the run reported **success**, not `skipped`, at the check level. So the validation guard can turn a check green without executing anything, which is worth stating in the PR body next to the caveat that already covers the skip: a green `claude-review` on any branch whose workflow differs from `main` means "did not run", and the history contains one such green already.

**2. Concurrency is not the cause — worth naming so nobody adds a guard expecting it to help.** I started down that road (many pushes across a five-PR stack, each spawning an Opus review on one shared token, instant `$0` failures being the classic rate-limit shape) and it does not survive the history: the 2026-08-18 20:17 ([32181438760](https://github.com/goodfire-ai/causalab/actions/runs/32181438760)) and 2026-08-19 13:22 ([32257735893](https://github.com/goodfire-ai/causalab/actions/runs/32257735893)) failures were isolated single runs with nothing else in flight, and they carry the same `num_turns: 1, total_cost_usd: 0, is_error: true` signature. A per-PR `concurrency` group would still be reasonable hygiene (`Tests` has one), but it belongs in its own change and it will not fix this.

Both entry points confirm the diagnostic step is the load-bearing part of this PR. The `claude.yml` smoke test's log also shows the tag-mode fetcher shelling out `git hash-object` for every file changed in the PR against a `fetch-depth: 1` default-branch checkout — dozens of `fatal: could not open 'causalab/neural/pytorch_hooks/…'` warnings before the failure. Non-fatal, but it means the action was reasoning about the PR from a tree that does not contain it, which is its own reason to want the execution log.

can-goodfire:
Agreed with the point already on this thread that a green `claude-review` here (and on #30's earlier attempt) is the action's workflow-validation self-skip, not evidence — that is in the PR body above, and it is why this cannot be verified before merge.

One datapoint that is *not* a skip, and that narrows what is left: the `@claude` mention I posted on #30 ran **`claude.yml` from `main`** — main's own copy, so it passed workflow validation and **actually executed**. It failed at the first turn with the same signature as every `claude-review` attempt:

```
"model": "claude-opus-5"          # init resolved --model opus
{"type":"result","subtype":"success","is_error":true,
 "duration_ms":2059,"num_turns":1,"total_cost_usd":0}
```
[run 32469715455](https://github.com/goodfire-ai/causalab/actions/runs/32469715455)

So a genuinely-executing run, with this token and this model pin, fails on its first API call. That excludes everything specific to the review workflow (prompt, `track_progress`, the inline-comment tool) and leaves exactly the two things the workflows share: the credential and `--model opus`. It does **not** decide between them — as the thread says, only a post-merge run can, and the diagnostic step here is what will name it. Worth knowing before merging that if the token is the cause, the fix is `claude setup-token` plus a secret update, not a code change.

can-goodfire:
Note on the `test` check here: this PR targets `main`, and `test.yml` checks out the PR head under `pull_request_target` — so it is running **main's pre-refactor suite**, not the tiered one from the protocol branch. It has been in `Run tests` for 25+ minutes (the same job takes ~2 min on protocol-refactor branches), and main's own last `test` run was a failure back in 2026-05-07. Whatever it reports is about main's suite, not about a two-line `.github/workflows` change — compare against main's own history before reading anything into it.

PR #41, opened 2026-08-26
Add pipelines/: the six-step research protocol, moved from silico

goodatticus:
Moves the causal-analysis workflow documentation out of the silico repo (`plugins/capabilities/causalab/`) and into causalab, next to the library it drives — and refactors the main flow while doing it.

Companion PR removing it from silico and repointing the skill: to follow.

## The new protocol

`pipelines/answer-research-question/` — six steps, each its own subdirectory:

1. **Behavioral analysis** — can the model do it, what does it get wrong, and (for multi-token generations) what kind of output is it producing
2. **Exploratory experimentation** — cheap probes for signal: logit lens, probing, ablation, steering, attribution, feature labels, plus single-pair interchange and residual-stream PCA
3. **Hypothesis generation** — the causal model, the competing alternatives, and the counterfactual datasets that separate them
4. **Hypothesis testing** — the hypothesis against its alternatives
5. **Generalize results** — how far the claim actually reaches
6. **Save results**

`answer-research-question.md` owns the shape of the whole thing: the flow chart, the roadmap doctrine, and the routing.

**The roadmap comes first.** Before any experiment, `ROADMAP_TEMPLATE.md` is copied into the working directory and filled in — what you expect from each step, what would shorten it, what would stop the project. It is revised at the end of every step, with an append-only log recording what actually happened and what in the plan is now wrong.

**Steps are phases, not experiments.** The outer document states the decomposition rule; each step document carries its own "Typical units of work".

**Routing out of hypothesis testing.** Strong positive result → generalize. Otherwise: back to hypothesis generation if the existing evidence supports another hypothesis worth proposing, or back to exploratory experimentation if it doesn't and the next hypothesis would just be a guess. With an explicit check first that the null isn't just an underpowered or broken test.

## What this replaces

`develop-causal-model/`, `explore-behavior/`, and `task-causal-analysis-pipeline/`. The first two are absorbed; the third was entirely retired Hydra orchestration (gate G2, viewer-spec merge, the plan/run/interpret chain) whose sequencing role the outer document now plays.

## Execution stubs

Those documents drove `scripts/run_exp.sh`, `causalab/analyses/`, `methods/` as Python, and SLURM dispatch — all removed on this branch in #20. The science is kept; the invocations are not. Every place a concrete command used to be carries:

> **Execution: stub.** What this needs, and why it isn't written yet.

Nine of them, indexed in `pipelines/README.md` with what each is waiting on. Notably: attribution needs new protocol vocabulary, not just a document; probing can't express its own loss because the metric vocabulary is closed over `lm_head` reads; feature labels have the `sae` featurizer but nothing on the labeling side.

`hypothesis-testing/` has **no** stub — it's written against the shipped method presets in `causalab/configs/methods/` and the `weekdays_8b` workflow.

## Moved unchanged

`implementation/`, `explore-subspace/`, and `subspace-causal-analysis-pipeline/` come across with a staleness banner and repointed cross-references. They're reference trees entered from within the steps, not steps.

## Also

- Silico-specific conventions stripped throughout — narration blocks, worker check-in gates, `~/.silico` paths. These documents are about using causalab now, not about one agent harness.
- `README.md` and `CLAUDE.md` point at `pipelines/`.
- `pyproject.toml` excludes `pipelines/**/templates` from ruff — the copy-from scaffolds carry `{{PLACEHOLDER}}` markers and are deliberately not valid Python.

## Verification

All relative links in the tree resolve (checked programmatically). `pre-commit run --all-files` passes.


goodatticus:
Changing direction — this documentation is staying in silico after all. Superseded by goodfire-ai/silico#6393, which keeps the tree at `plugins/capabilities/causalab/` and applies the same six-step refactor there.

PR #45, opened 2026-08-27
Workflow protocol v2: two step types, one reference grammar

can-goodfire:
Stacked on #20 (`can/protocol-refactor`), the same base #39 and #42/#44 use.

Collapses the workflow vocabulary from four closed step types (`protocol` / `transform` / `select` / `plot` over a versioned op registry) to **two**: `protocol` stays declarative because that is where load-time bite lives, and `script` is *inputs → one Python script → declared outputs*.

**This PR partly reverts #39** (the `transform` step type, already merged into the base). Saying so up front rather than letting a reviewer find it: the registry's numerics survive as shipped scripts with their oracle tests unchanged; what goes is the registry itself.

## Why

The registry admitted ops only *by pull request*, so a one-off corpus-mean harvest — or the harvest-and-difference that produces a steering direction — was expressible nowhere. Not a protocol document (it touches no network through the intervention vocabulary), not a registered op. That work was happening outside the record, which is the failure the workflow document exists to prevent. Of the nine execution stubs in silico's `pipelines/README.md`, four are closed outright by a script step (steering direction, mean-ablation harvest, logit lens, probing loss), two partly, and three are correctly out of scope (attribution needs protocol vocabulary; feature labels and `develop_hypothesis` stay silico's).

`select` and `plot` survive as **shipped scripts** (`causalab:select`, `causalab:plot`), so the common reductions stay one line to author.

## Two commits

1. **`JSON and safetensors as the only two formats`** — metric tables become native JSON (an array of row objects). Labels repeat; that is the deliberate trade for a file `jq` and a human can both read. Re-pins every corpus/golden/workflow digest by construction (`file_path` is in the canonical form), so it is a pure-rename commit reviewable by eye. `pyarrow` stops being a direct dependency.
2. **`Workflow protocol v2`** — the vocabulary change.

## What changed in the contract

- **One reference grammar** (spec §3): a **locator** — `{"path": P}` (absolute if leading `/`, else repo-relative) or `{"step": S, "file": F}` — plus an optional **selector**: `key` for JSON, `entry` for safetensors. Seven forms became 2 × 3, and `file`/`artifact`/`document` (three tags for three roots) became one `path`. `file` now always means a name a step *declares*; `path` a location the workflow does not own.
- **No `save` section, no sink rule** — the run tree *is* the publication, so the runner's copy loop is gone and a terminal plot step needs no blessing. Cost, stated: nothing flags a step whose outputs nobody reads.
- **`output_dir`** names the workflow's own directory, the CLI supplies the root (`--out`), so documents stay free of absolute paths. Excluded from the digest: it names *where*, not *what*.
- **Script content hashes are in the digest**, which is what makes `--resume` correct. Hashing needs no import, so `validate`/`digest` stay torch-free — rule 6 parses and hashes and *never* imports.
- **`is_deterministic`** per step; **`runtime.isolate`** runs a step in a subprocess with its own deps.

## Four decisions the implementation forced — please look at these

1. **`outputs.<slot>.keys` is new vocabulary.** With `select` a script, the loader lost the `emit` table it used to build representative values for step-dependent inner documents, which makes the weekdays pipeline unvalidatable. A `.json` output now declares `keys` mapping each emitted name to a **representative value** — a value, not a dtype, because `positions.tap` is a position spec and `"string"` would not type-check as one. Mutually exclusive with `columns`.
2. **A `.png` is neither of the two formats.** Resolved in the format rule's favour: `causalab:plot` declares the *aggregated table it plotted* and writes the image beside it. Better provenance than the picture — the numbers are recomputable and referenceable.
3. **A script can declare identity fields only it knows.** Dropping the registry dropped `identity_from_params`; the first run refused with `[V15] ArtifactIdentity is missing 'k'`. `write_tensor(..., identity={"k": k})` is that channel. Inherited fields win on conflict: what the inputs *prove* beats what a script claims.
4. **The as-written aggregation rule is data-driven.** v1 exempted a `transform` table *by producer type*. With no types left the discriminator is the data: group by published axes; else if an `example` column is present the table is one group; else the rows are already the unit. `select` and `plot` share one function so a figure and a chosen value can never disagree about what a row is.

## Load-time → run-time, honestly

Column checks, plot axis coverage and the op-record checks are now step-time refusals. Mitigated but not erased: because `protocol` stays declarative the runner still knows each protocol step's sweep axes, and publishes them in a per-step `_step.json` that the shipped scripts read. Group-by-coordinates-then-mean survives as *behaviour* rather than derived magic in the document model.

## Verification

- **1689 passed** (baseline 1692 — net test-count change from the rewrite, no skips), 8 deselected, ~15 s CPU. 20 CPU structural goldens. `pre-commit run --all-files` clean.
- The **weekdays capstone** runs the same pipeline shape end to end on tiny-random through the real CLI — locate → select → DAS fit → apply → plots — and now also pins that **editing a step's script busts `--resume`**, the concrete payoff of the hash-in-digest decision.
- `causalab validate` of a workflow whose script does `import torch` leaves torch out of `sys.modules` (subprocess-checked), so parse-and-hash really is import-free.
- Isolation asserted by **pid comparison**, so it cannot silently regress to the in-process path.
- The three op oracles ported **unchanged** — the numerics did not move when the ops left the registry.

## Size — the estimate was wrong

The plan predicted ≈−1,100 source lines. Actual: **2,783 → 2,935 (+152)**. `workflow.py` shrank by 32 lines, not 580 — deleting the registry's record machinery freed almost exactly what the reference grammar, output declarations, script hashing and `runtime` parsing then consumed. 135 of the added lines are capability that did not exist before (`harvest_difference`, the isolation shim).

What did shrink is the number of things to know: step types 4→2, wiring mechanisms 3→1, ref forms 7→2×3, document sections 4→3, storage formats 3→2. Simplification of the *contract*, not the line count.

## Known follow-ups

- Editing a shipped script moves `workflow_digests.json` (its hash is in the digest). Correct, but builtin edits now carry a re-pin.
- A `set` block still authors IM-spec refs (`{"artifact": …}`) while the step's own `inputs` use the §3 grammar — the two grammars meet in exactly one place. Aligning them is deliberately out of scope here.
- `docs/test_migration.md` not updated; its retired-module ledger predates this change.
- #20's description documents the v1 workflow layer at length and goes stale when this lands.


can-goodfire:
## Review round 1 — all four items done (`f3afa1d`)

**1. `protocol` → `intervention_protocol`.** `"type": "protocol"` sat next to "a protocol document", "the protocol layer" and `causalab/protocol/`. The step table is the one place the type is written, so the long name costs nothing.

**2. `.png` / `.pdf` / `.html` are legal outputs**, documented as *visualization* formats: legal to declare, but they carry no record — no `columns`/`keys`, no shape check, no ArtifactIdentity stamp, existence is the whole contract. **`.png` is the default and preferred over `.pdf`** unless a document asks for pdf explicitly.

That preference is not restated anywhere: it's `causalab.io.plots.figure_format.normalize_figure_format(value, default="png")`, which **already existed** and which the renderer now calls. `.html` needs an interactive renderer, so the matplotlib script refuses it by name and points at the plotly idiom in `io/plots/distance_plots.py` rather than letting `savefig` raise about supported formats.

This **reverses** my earlier resolution (declare the plotted table, write the image beside it). The figure is a declared output again; the table survives as an *optional* second output, because declaring the rows behind a picture is what makes it checkable.

**3. `steps/` is gone.** It was a catch-all with a tidier name — science does not belong inside the workflow package. Scripts are now filed by subject:

| was | now | why |
|---|---|---|
| `steps/builtin/{fit_pca,harvest_difference,head_stats,paired_ttest}.py` | **`causalab/analysis/`** | numerical analysis — fits, statistics, intervention operands. Importable without the workflow layer |
| `steps/builtin/plot.py` | **`causalab/io/plots/workflow_figures.py`** | rendering, beside the other 17 plot modules |
| `steps/io.py` | **`causalab/io/step_io.py`** | IO, where IO lives |
| `steps/builtin/_sidecar.py` | **`causalab/io/step_record.py`** | the `_step.json` format: writer, reader, shared aggregation rule |
| `steps/_shim.py` | **`causalab/workflow/isolate.py`** | subprocess entry — genuinely the runner's |
| `steps/builtin/select.py` | **`causalab/workflow/scripts/select.py`** | the one script whose purpose *is* wiring steps |

Two things this turned up. My `plot.py` had been **reimplementing a heatmap** that `io/plots/utils.py::create_heatmap` already provides, and reinventing the png/pdf default that `figure_format.py` already implements — both consequences of putting it in a package where the existing plotting code wasn't visible. And `step_record` *had* to move to `io/`: `io/plots/workflow_figures` reads it while the runner reads `io/step_io`, so leaving it in `workflow/` was an `io ↔ workflow` cycle. The layering guard caught it and now asserts the new direction.

**`script` is a locator**, mirroring the `inputs` grammar:

```json
"script": {"module": "causalab.analysis.fit_pca"}
"script": {"path": "scripts/probe.py"}
```

`importlib.util.find_spec` resolves a dotted name to a file *without executing it*, so rule 6 still parses and hashes and never imports. The old `causalab:<name>` namespace needed a registry — the thing this PR removes — and hid which code ran behind a lookup. A module path says it, and a script can live where it belongs by subject instead of where a resolver happens to search.

**4. `protocol/` and `workflow/` fully separated.** Your preference, and it's now an enforced invariant rather than a convention:

- `protocol/workflow.py` → `workflow/document.py`
- CLI splits three ways: `causalab/cli.py` owns argument parsing and dispatches on the `steps` section; `protocol/cli.py` and `workflow/cli.py` own their own verbs. Entry point is `causalab.cli:main`.
- `test_protocol_does_not_link_against_the_workflow_layer` asserts `protocol/` imports nothing from `workflow/`, `analysis/` or `io/` — so the intervention protocol really is usable on its own.
- Dependency runs `workflow/` → `io/` → `protocol/`, one way.

### Found while doing this

The workflow CLI's model pre-registration checked `type != "protocol"`. After the rename it silently skipped every step, so **every workflow run failed** on an unregistered model. Caught by the corpus tests, not by inspection — worth knowing the rename had teeth.

### Verification

1705 passed, 8 deselected; `pre-commit run --all-files` clean; `causalab validate`/`explain` on the shipped `weekdays_8b.json` give the same 5-level schedule and the same 64/9/1 point counts.

Still open from the original description: the two ref grammars meeting inside a `set` block, and `docs/test_migration.md` untouched.

PR #51, opened 2026-08-27
top_k over any read, with a mandatory ranking rule

can-goodfire:
## What

`top_k` works on **any** read — `block_output`, `mlp_activation`, a featurizer output — not just `lm_head`.

Before this, `validate.py` refused a `top_k` whose `of` read tapped anything but `lm_head`, and `requires()` charged `full_logits` to every document containing a `top_k` regardless of what it read. That made the one reduction that keeps a wide tensor off disk unavailable exactly where the tensor is widest: a 4k-dim residual stream, or an SAE/BSF featurizer output at 100k+ latents, times every example and every position, had to be saved whole and argsorted afterwards when the author only wanted the top few entries per row. Like `reduce: mean` on a save entry (§2.12), the reduction should happen where the rows are gathered.

Per the owner's call this stays **one overloaded `top_k` kind** — no `top_dims` / `top_blocks` / `top_features` siblings. The semantics are disambiguated by mandatory parameters.

### 1. Ranking rule — `by`, mandatory, `value` | `abs_value` | `prob`

Exact spelling: field **`by`** in `METRIC_FIELDS["top_k"] = ("k", "by")`; enum `TOP_K_RANKINGS = ("value", "abs_value", "prob")` in `causalab/protocol/schema.py`.

| `by` | ranks on | legal on |
|---|---|---|
| `value` | the k largest **signed** entries | any read |
| `abs_value` | the k largest by `\|x\|`; the reported value stays signed | any read |
| `prob` | softmax the last axis, then the k largest probabilities | **`lm_head` reads only** |

Mandatory rather than defaulted, because only the author knows what the axis is and the answers genuinely differ: a vocabulary projection has no meaningful negative entries for top-k purposes, while a residual stream and a signed feature code do. Ranking a 100k-latent SAE code by signed value and by magnitude return *different sets*, and defaulting silently picks one — the failure mode being a plot that means something other than its caption. `by` is also not sweepable (§3 wrappers refused), for the same reason `token_form` isn't: it forks a campaign on how a result is read, not on a research variable.

Rather than add a *third* concept for normalization, `prob` folds it into the ranking enum — softmaxing and then ranking by probability is one coherent operation, and it is exactly the pre-change behaviour. Validation refuses `prob` off `lm_head`: a softmax across neurons or SAE latents normalizes over an axis that is not an event space, so the "probabilities" would be probabilities of nothing.

### 2. Output schema — four columns, each with one fixed identity

| column | meaning | emitted when |
|---|---|---|
| `indices` | index along the read's last axis (token id on `lm_head`, neuron on `mlp_activation`, latent on a featurizer output) | always |
| `tokens` | that index decoded as a token string | the read taps `lm_head` |
| `values` | the **raw** read value at that index | always |
| `probs` | the softmax probability over the vocabulary | `by: "prob"` |

A column is **absent** rather than reinterpreted, so no column ever changes identity between documents. In particular `values` is always the raw value — a *logit* under `by: "prob"`, not the probability — so a downstream reader never has to know the ranking rule to know what it is holding; the normalized number lives in its own column. `tokens` follows the read and `probs` follows `by`, independently: `by: value` on an `lm_head` read still decodes (indices really are token ids) but emits no `probs`.

`k` is now range-checked against the read width (`1 ≤ k ≤ width`) instead of letting `torch.topk` raise.

### 3. Capability derivation

`full_logits` is now derived from the `of` read's component, not from the metric kind alone (`causalab/protocol/backend.py`). A `top_k` over a `block_output` obliges no vocabulary projection anywhere and must not route the document onto a full-vocab backend. `class_probs` is unaffected in practice — validation still binds it to `lm_head`, so it always derives `full_logits` as before.

The same predicate is needed in three places (capability derivation, validation, and the reduction, which has to know whether an index is worth decoding), so it is one shared function: `schema.metric_reads_vocabulary(doc, metric)`. Three copies of "is this read `lm_head`?" that could drift is exactly the bug class this avoids.

### 4. `token_form` — stays refused on `top_k`, correctly

Confirmed still right under the new semantics. `token_form` governs how an *authored string* becomes a token id. `top_k` authors no strings: it reports indices it found and decodes them only as a courtesy when the read happens to tap `lm_head`. There is nothing for the knob to apply to. Docstrings and §2.10 prose updated to say why in the new framing rather than the old "decodes ids it found".

### 5. Siblings not loosened

The `lm_head` guard is now keyed on `_ANY_READ_METRIC_KINDS = {"kl", "top_k"}` — `kl` was already exempt (it compares two reads against each other), `top_k` joins it. `logit_diff`, `token_logit`, `cross_entropy`, `class_probs` and `match` all still refuse a non-`lm_head` read, and there is a parametrized test pinning each one.

## Corpus / digest re-pins

Making `by` mandatory breaks any existing document using `top_k`. Two corpus documents do — `11_probe_generate_im.json` and `12_probe_variable_im.json` — and both rank logits, so both gained `"by": "prob"`, preserving their behaviour exactly. `tests/protocol/corpus_digests.json` re-pinned via `tests/protocol/update_corpus_digests.py`; **only those two entries moved**, and only because their canonical form gained one field. `workflow_digests.json` and `golden_digests.json` needed no re-pin (their tests are green untouched) — nothing in `tests/golden/` or the workflow corpus uses `top_k`.

⚠️ **One thing I had to touch outside the brief.** `update_corpus_digests.py` loads *every* corpus document, and `12_probe_variable_im.json` currently fails V10 ("a metric's file_path must end in '.json'") because its two metric save paths still say `.parquet` — a pre-existing #44/#45 semantic merge conflict on this base. The script aborts on it, so **re-pinning was impossible without fixing it**. I changed those two paths to `.json` and switched `test_run_corpus.py`'s two `pd.read_parquet` calls to `table_frame`, the same helper its sibling `test_11_...` already uses. This is the minimum needed to unblock the re-pin; I did not touch the other family of pre-existing failures.

## Test evidence

Baseline on a pristine checkout of `can/protocol-refactor` @ 6e467ea: **17 failed, 1721 passed, 8 deselected** (pre-existing #44/#45 merge artifacts + two transformers-5.16 sentencepiece pins).

This branch: **11 failed, 1748 passed, 8 deselected**.

- **New failures: zero.** The 11 are a strict subset of the baseline 17.
- **6 baseline failures fixed** as a side effect of the forced corpus-12 fix: the 4 `test_corpus.py[12_probe_variable_im.json]` params, `test_run_corpus.py::test_explain_and_digest_work_on_every_corpus_file`, and `test_run_corpus.py::test_12_probe_variable_scores_every_step_and_reports_what_was_said`.
- **+21 new tests**, all passing.

New coverage:

- `tests/protocol/test_schema.py` (`unit`) — `by` is mandatory (with a message that tells a pre-`by` author their document meant `prob`), the enum is closed, all three members parse, and `by` is not sweepable.
- `tests/protocol/test_validation_rules.py` (`unit`) — `top_k` binds to a read at any component; `by: prob` off `lm_head` is refused; `by: prob` on `lm_head` is legal; and a parametrized case pinning that `class_probs` / `token_logit` / `cross_entropy` / `match` are *still* `lm_head`-only.
- `tests/protocol/test_backend_routing.py` (`unit`) — `top_k` over an `lm_head` read still needs `full_logits`; `top_k` over a `block_output` does not.
- `tests/neural/pytorch_hooks/test_metrics.py` (`numerical_unit`) — hand-computed oracles against a signed 1×6 code `[0.5, -7.0, 3.0, -0.25, 6.0, -2.0]`, deliberately built so the largest magnitude sits on a *negative* entry and the two rules cannot agree: `by: value` → indices `[4, 2, 0]` / values `[6.0, 3.0, 0.5]`; `by: abs_value` → indices `[1, 4, 2]` / values `[-7.0, 6.0, 3.0]` (sign preserved). Plus column presence per branch, per-row independence on a 2-row batch, and `k` out of range on both ends.
- `tests/neural/pytorch_hooks/test_run_corpus.py` — corpus 11 end-to-end now pins the full four-column shape.

`uv run --frozen pre-commit run --all-files`: green (ruff check + ruff format).

## Docs

`docs/intervention_protocol.md` §2.10 in the same commit: the metric table row, a new "`top_k` — one kind over any read" subsection carrying the `by` and column tables with their rationale, a "Reads a kind may bind to" paragraph, the domain table's wording (`distribution` consumes "the read's dense value", the vocabulary projection for every kind but `top_k`), the `token_form` bullet, and §8's `full_logits` row.


can-goodfire:
CI `test` is red, and it is red **exactly** as the base branch is. The 11 failures the job reports are a strict subset of the 17 on a pristine `can/protocol-refactor` @ 6e467ea — no new failure, and 6 of the baseline 17 are fixed here (the corpus-12 ones, unblocked by the forced `.parquet` → `.json` fix described in the body).

Remaining, all pre-existing:
- 8 × `test_generate_metrics.py` — #44/#45 merge artifact, those docs still save metrics to `.parquet` (V10). Same root cause as corpus 12, but in test-local documents; left alone to avoid entangling with this change.
- 1 × `test_validation_rules.py::test_decode_over_a_continuation_read_is_legal` — same `.parquet` cause.
- 2 × `test_metrics.py` (`test_a_pinned_form_refuses_rather_than_falling_back`, `test_first_token_skips_the_lone_space_piece`) — pin legacy sentencepiece behaviour the transformers 5.16 bump removed. Unrelated.

`lint` is green.

can-goodfire:
Review follow-up, addressed in 5d785ec: `metric_reads_vocabulary` now also requires the read to be plain (no `featurizer`, no `dims`) — the site alone waved a featurized lm_head read through `by: "prob"` (softmax over latents), decoded latent indices as token strings, and let token-space kinds bind to a read their resolved ids mis-index. Rule 4 routes both vocabulary checks through the predicate and names which transform broke the axis. Capability derivation deliberately does *not* reuse it: a featurized lm_head read still obliges the full projection (the featurizer consumes it), so `requires()` gets its own `_metric_read_obliges_full_projection` — the two questions diverge exactly on transformed lm_head reads. Suite: 1766 passed, 0 failed.

PR #53, opened 2026-08-27
PR4 — attention_probs: read the pattern, and make a write to it actually land

can-goodfire:
PR4, the last of round 1. **Stacked on #52** (review #48 → #49 → #50 → #52 first).

Reading the attention pattern is an ordinary tap — `self_attn` returns `(attn_output, attn_weights)`, so it is element 1, and 📐 on the fixture's one full-attention layer it is `(1, 8, 6, 6)` with rows summing to 1. **Writing it is the whole of the work**, for a reason worth stating precisely because it looks like it shouldn't be one.

## Why a forward hook cannot write the pattern

```python
attn_output, attn_weights = attention_interface(self, q, k, v, mask, ...)
...
return attn_output, attn_weights
```

By the time the mixer returns, `attn_output` has already been computed from the pattern **inside** the attention function. A `register_forward_hook` can rewrite element 1 of the returned tuple and change nothing — the same silent-no-op shape as writing `router_logits` in #52, and measured the same way: **0.0** change in the logits. The edit has to happen between the softmax and the value multiply.

transformers resolves that function per forward via `ALL_ATTENTION_FUNCTIONS.get_interface(impl, eager_attention_forward)`. 📐 **`"eager"` is not registered by default** — the registry holds `sdpa`, the flash variants, `flex_attention` and the `paged|*` family — so that call falls through to the module's own implementation. Registering `"eager"` therefore *inserts* a wrapper rather than replacing one, and removing the key restores the original exactly. The backend already forces eager at load, so this is the only implementation to wrap.

## What makes duplicating library internals safe here

The wrapper calls the real eager function, applies the edit, then redoes only the two lines after the softmax:

```python
value_states = repeat_kv(value, module.num_key_value_groups)
attn_output = torch.matmul(edited, value_states).transpose(1, 2).contiguous()
```

Duplicating internals is normally how a backend rots silently. So it is pinned: **with an edit that changes nothing, the recomputed logits must equal the unpatched logits exactly** — 📐 measured max difference 0.0. A future transformers that changes the post-softmax math (different scaling, a dropout that isn't a no-op in eval, a different output reshape) fails that test instead of drifting quietly.

## Scope — and what is deliberately refused

This tensor has **two position axes**, and its feature axis IS a position axis. So `_gather` (dim 0 batch, dim 1 position) would index *heads* with position indices, and `dims` would slice *key positions* as though they were features. Both produce plausible numbers from the wrong tensor.

Round 1 exposes the whole pattern at `pos: "all"`. Each form that needs a typed feature-shape descriptor is refused, by name:

| form | outcome |
|---|---|
| `pos: "all"` read | ✅ `(b, H, s, s)` |
| `pos: {index: …}` / `{span: …}` | ❌ refused — names follow-up F1 |
| `featurizer` | ❌ refused — F1 |
| `dims` | ❌ refused — F1 |
| whole-pattern `swap` write | ✅ logits move |
| write at a DeltaNet layer | ❌ refused on the architecture (#50's stream check) |
| interchange across different lengths | ❌ refused (not broadcastable) |

A write replaces the whole pattern, which is what an interchange on attention means and what nnterp's own check does (`self[layer] = rnd`).

## `layout` gains `"native"` — not a fourth conversion case

§6.2 review point 3 asked PR4 not to bolt a fourth case onto `layout`. `"native"` claims **nothing** about the axes and converts by doing nothing. It exists so this tap cannot be mistaken for the contract by defaulting to `"bsd"` — where the conversions are also the identity, but the *claim* would be false. There is no conversion logic for F1 to unpick; F1 replaces the marker. The layout suite's generic assertions now run over `CONVERTING_LAYOUTS`, since "reaches the `(batch, position, feature)` contract" is not a statement about `"native"`.

## nnterp's three-part check, ported

Against real parsed-and-validated documents: shape `(b, H, s, s)`, rows sum to 1, and **writing changes the logits**. Plus registry hygiene (restored after a run, after an exception, and never installed when a run has no pattern writes) and scoping (a mixer absent from the edit map is untouched).

`writable_attention_probs` is now declared, so §8 routing accepts these documents.

**One of #50's tests is discharged rather than deleted:** `attention_probs` at a full-attention layer used to assert `NotImplementedError` — a roadmap statement. It now asserts the tap resolves, while the asymmetry that matters (a DeltaNet layer refuses *permanently*) stays pinned.

## Gates

- `pytest -m "not golden"` — **1856 passed, 8 deselected**, 0 failed
- `pre-commit run --all-files` — green repo-wide
- parity goldens and corpus digests **untouched**
- 16 new tests

Plan note: `notes/causalab-hookpoint-vocabulary-plan.md` §6 PR4. **Round 1 is complete with this PR.**


can-goodfire:
Review follow-up, addressed in cc76483: the eager wrapper hardcoded `modeling_qwen3_5_moe`'s `eager_attention_forward`/`repeat_kv` for every module while the registry entry intercepts *every* attention forward — a family whose eager math differs (gemma-2's soft-capping) would have been silently replaced, unpinned. `_post_softmax_math` now resolves both from the mixer's own modeling file per call, refusing by name when a family exports neither; pinned by an identity bit-exactness + causal write test on tiny-llama. Also: the whole-pattern write branch read `write.do.payload` unconditionally, so non-swap mechanisms are now refused by name (round 1 is interchange; F1 owns re-weighting). Suite: 1861 passed, 0 failed.

PR #65, opened 2026-08-28
One runnable protocol per method, on the hybrid A3B tower

can-goodfire:
Stacked on #20.

The Silico causalab documents (goodfire-ai/silico#6393) describe six exploratory and testing methods and mark each one **Execution: stub** — the Hydra runner that used to invoke them was retired by the protocol refactor and nothing replaced the invocations. This adds them back as protocol documents, pointed at [`Qwen/Qwen3.6-35B-A3B`](https://huggingface.co/Qwen/Qwen3.6-35B-A3B) so they are exercised against a hybrid MoE tower rather than a dense Llama.

## What lands

Thirteen documents under `causalab/configs/protocols/qwen3_6_a3b/`, two workflows, the `weekdays` tables they run on, a registry entry for the model, and a shape test.

| document | method | points |
|---|---|---|
| `probe` | step 2 · probe | 1 |
| `logit_lens` | step 2 · logit lens (layer × position) | 77 |
| `locate_scan` | step 2 · paired examples / step 4 · locate | 40 |
| `knockout_head` | step 2 · knockout, attention head grid | 160 |
| `knockout_mlp` · `knockout_mlp_band3` | step 2 · knockout, MLP bands | 40 · 1 |
| `harvest` | step 2 · PCA's input + the mean-ablation reference | 1 |
| `head_attribution` | round 2's derived `attention_result` | 1 |
| `interchange` · `control_positive` · `control_negative` | step 4 · the test and its two controls | 1 each |
| `das_sweep` · `das_apply` | step 4 · DAS over k × seed, then the winner held out | 6 · 1 |

`qwen3_6_a3b_exploration.json` fans step 2 out; `qwen3_6_a3b_weekdays.json` chains step 4 (locate → select → fit → select → apply → plot).

PCA needed no new code — `causalab.analysis.fit_pca` already ships and is already named in the workflow spec's script table; what was missing was a document wiring a harvested read into it.

## The registry entry is load-bearing

The pure verbs are deliberately registry-only so digests never depend on the network (`cli.py:194`), which means `validate` and `explain` — the two verbs the Silico hypothesis-testing document tells an agent to reach for *first* — refuse outright on a model that is not registered. Shipping configs that name a model without registering it would have made that advice fail on the target model.

## Three things the documents encode rather than assume

Each was found by running them, not by reading:

- **A prompt-frame metric reduces exactly one position per example.** So the logit lens's (layer × position) grid is two swept axes, not one `pos: "all"` read that `top_k` then refuses.
- **`{"variable": "x"}` is looked up per role.** A bare `subject` column resolves against the counterfactual text too and refuses there; the tables carry `input_variables` / `counterfactual_inputs_variables` instead.
- **All seven weekday answers are ambiguous under this tokenizer** — both the bare and space-prefixed forms are single tokens. `token_form: "auto"` warns and resolves by luck, so every metric naming a token string pins `space_prefixed`.

## What the tower's shape costs

`full_attention_interval` is 4, so of 40 layers only **ten** carry a full-attention mixer. The head grid is 10 × 16, not 40 × 16; `mlp_activation` is not addressable at all (every layer is a sparse-MoE block, which has no `act_fn`); and KV-space components have two heads against sixteen query heads, so an out-of-range `head` is refused rather than silently yielding an empty slice.

## Verification

`2249 passed` (`-m "not golden"`), lint clean. Every document was smoke-run against `tiny-random/qwen3.5-moe` — the same architecture class, `Qwen3_5MoeForConditionalGeneration` with the same 3:1 linear:full layer types — and both workflows complete there. `sum_h attention_result == attention_output` holds to 1.0e-7.

The run on the checkpoint itself is in flight; results follow in a comment.

## Follow-up

These are written to become the golden tier: deterministic, small, complete `save` manifests. What they are pinned on here is *shape* — point counts and which layers may carry an attention component — because the numbers need the 67 GB checkpoint.

### DIFF
--- a/causalab/configs/data/weekdays/build.py
+++ b/causalab/configs/data/weekdays/build.py
@@ -1,0 +1,43 @@
+"""Emit the weekdays tables. Fully enumerated, so there is no seed to record:
+7 base days x the counterfactual offsets assigned to each split."""
+
+import json
+import sys
+import pathlib
+
+DAYS = ["Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday", "Sunday"]
+TEMPLATE = "If today is {day}, tomorrow is"
+
+
+def rows(offsets):
+    out = []
+    for k in offsets:
+        for i, day in enumerate(DAYS):
+            cf_day = DAYS[(i + k) % 7]
+            out.append(
+                {
+                    "input": TEMPLATE.format(day=day),
+                    "counterfactual_inputs": [TEMPLATE.format(day=cf_day)],
+                    "answer": " " + DAYS[(i + 1) % 7],
+                    "base_answer": " " + DAYS[(i + 1) % 7],
+                    "cf_answer": " " + DAYS[(i + k + 1) % 7],
+                    "label": " " + DAYS[(i + k + 1) % 7],
+                    # Per-role prompt variables. A bare `subject` column would
+                    # be looked up against BOTH roles' texts and the
+                    # counterfactual names a different day, so a
+                    # {"variable": "subject"} position would refuse on it. The
+                    # `<field>_variables` sibling is the per-role spelling:
+                    # `input_variables` for the base column, and a list for the
+                    # list-valued counterfactual column, indexed the same way
+                    # the field is.
+                    "input_variables": {"subject": day},
+                    "counterfactual_inputs_variables": [{"subject": cf_day}],
+                }
+            )
+    return out
+
+
+root = pathlib.Path(sys.argv[1])
+for name, offsets in (("train", (1, 3)), ("test", (2,))):
+    (root / f"{name}.json").write_text(json.dumps(rows(offsets), indent=2) + "\n")
+    print(name, len(rows(offsets)), "rows")
--- a/causalab/configs/data/weekdays/test.json
+++ b/causalab/configs/data/weekdays/test.json
@@ -1,0 +1,128 @@
+[
+  {
+    "input": "If today is Monday, tomorrow is",
+    "counterfactual_inputs": [
+      "If today is Wednesday, tomorrow is"
+    ],
+    "answer": " Tuesday",
+    "base_answer": " Tuesday",
+    "cf_answer": " Thursday",
+    "label": " Thursday",
+    "input_variables": {
+      "subject": "Monday"
+    },
+    "counterfactual_inputs_variables": [
+      {
+        "subject": "Wednesday"
+      }
+    ]
+  },
+  {
+    "input": "If today is Tuesday, tomorrow is",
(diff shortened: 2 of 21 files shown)


can-goodfire:
## Run on the checkpoint — job 1426492, 1× H100, ~10 min wall

Every document and both workflows executed on `Qwen/Qwen3.6-35B-A3B`. Exit 0 throughout; artifacts under `/mnt/data/artifacts/can.rager/a3b-protocols/1426492`.

The point of running them was not the exit codes — it was whether the numbers hold together. They do, and two independent methods agree.

### The controls bracket the pipeline

| | IIA |
|---|---|
| positive control — transplant `block_output` @ L39 | **1.000** |
| negative control — same swap at prompt position 0 | **0.000** |
| probe — model answers the task | **14/14** |

### The scan localizes, and the lens confirms it independently

The 40-point locate scan puts the weekday at the **subject token from layer 10 onward** (IIA 1.000) and at the **answer slot only at layers 36 and 38**.

The logit lens — a different mechanism entirely, a residual transplant into the last block — has the answer first decoding at the answer slot at **layer 36**. Same depth, no shared machinery. The L39 identity point reproduces the model's own prediction (`" Tuesday"`, p=0.70), which is the built-in check that the lens is wired right.

### That pairing is why `interchange.json` changed

The shipped example originally tested the answer slot at L20 and scored **IIA 0.000 / logit_diff −4.83**. Not a broken document — the wrong cell. Same layer, subject token instead: **IIA 1.000 / logit_diff +5.13** (confirmed in job 1426498 on `1d463e1`). A sign flip from moving the position, which is the argument for locating before testing, as data.

### Head knockout

160 cells over the ten full-attention layers. Baseline answer logit 18.15; **L39 head 13** is the most necessary single head at **15.54**. The 48 s it took is the whole grid.

### The DAS chain

`locate → best → fit → best_fit → apply` plus both figures, end to end:

- `best` selected **layer 36, answer slot** — one of the two cells the scan flagged
- `fit` over k × seed: **k=4 → −3.76, k=8 → −3.08, k=16 → −0.83** (mean train logit_diff). Monotone in rank — the "a higher-rank subspace always fits better" caveat, visible in the shipped data
- `best_fit` → k=16, seed=0; `apply` on held-out: −1.38, match 1/7. A real train/eval gap rather than a number to celebrate

### Cost

A 67 GB bf16 load per invocation (~20 s) dominates. The 40-point scan took 36 s, the 77-point lens 47 s, the 160-point head grid 48 s.

can-goodfire:
THese configs were mainly for demo purposes. The pipeline skill from got superseeded by causalab/research, so we're simply discarding these demos now

PR #75, opened 2026-08-31
Demos as markdown around runnable documents

can-goodfire:
Stacked on **#20** (`can/protocol-refactor`). Read that first — this PR is what its runner made possible.

`main` has `demos/onboarding_tutorial/*.ipynb` and `demos/weekdays_geometry.ipynb`; #20 has no demos at all, and its README says the notebook demos were retired. This restores them in the shape the refactor argues for: a notebook existed to carry its own execution, `causalab run` now does that, so what is left is the experiment.

## The format — `docs/demos.md`

Seven sections in a fixed order, a mandatory header table, and one rule that carries the rest: **`Results` has exactly one subsection per numbered question in `Experimental design`**, same order, same heading. A question without an answer becomes a stated gap instead of a silent omission.

Two choices in it are deliberate rather than conventional:

- **The document comes before the experimental design.** The document *is* the thesis — reading it should be enough to know what ran — and the design section exists to explain its choices. Inverting the usual order is what stops the prose restating JSON fields.
- **Every header carries a `Reproduced` field** that is either `✓ <date, engine, digest>` or `⚠ <what is stale>`, and nothing else. A demo whose figures predate its documents is useful; a demo that hides it is not.

§5 (*Voice*) is a characterization of the notebooks this replaces — person and tense, what bold and italics are each for, "say the number in the sentence", failure modes as content, and a *Not this* table. The tone was already right; the point is to make it repeatable.

## The four demos

| demo | question | shape |
|---|---|---|
| `01_define` | does this counterfactual dataset tell two variables apart? | dataset design, no network, **CPU, reproduced today** |
| `02_trace` | for one pair, which cells carry the answer symbol? | 16 layers × 22 positions, 352 points |
| `03_localize` | across 64 pairs, which cell carries it most reliably? | 128 points, as a document *and* a 3-step workflow |
| `weekdays_geometry` | where and how is the answer day represented? | the workflow exemplar: 4 RQs, 10 steps, 5 derived levels |

Each carries its own `protocols/`, `workflows/`, `data/` (tables **plus** their manifest sidecars) and `figures/`. Every document validates with `--data`; every `explain` block in the markdown is pasted output.

The arc is deliberate: 01 shows that a crafted counterfactual design drives the `answer`/`answer_position` confound to 0.000, 03 spends that (pairwise `match` is the right metric because *the dataset* deconfounded — the pre-refactor "centroid mode" existed only because the generator did not), and `weekdays_geometry` shows what it costs when a task ships only a random generator: 3 pairs in 64 are literally the same prompt, 14 share an answer, so the IIA floor is 0.219 and the entity/number columns are uninterpretable as localization.

### On the figures, plainly

Every figure is carried from the **pre-refactor reference runs** — the documents here have not been run on a GPU. Each demo's header says so, and each caption says exactly how the reference differs from the document above it (different pair, centroid vs pairwise scoring, entity-only vs independent resampling, layer 28 vs the pinned 18). `Results` subsections with no number say "no result" and name what would produce it. Regenerating them is the obvious follow-up and turns every ⚠ into a ✓ and a digest.

Two things surfaced while writing, recorded in `Limits` rather than fixed here:

- The source `weekdays_geometry.ipynb`'s prose says layer 18 while its config says 28; the PCA/manifold/steering figures are the config's.
- `causalab/configs/protocols/weekdays_locate_scan.json` and `harvest.json` address `{"variable": "subject"}`, a column the `natural_domains_arithmetic` serializer never emits (it emits `entity`). Those presets resolve only against the hand-made test fixture. The demo carries its own documents, so it is unaffected.

---

## Changes outside `demos/` and `docs/`

Three, each one something the demos need in order to be checkable rather than decorative.

### 1. `tests/demos/test_demos.py` — new, `unit` tier, 39 tests in 0.7 s

A demo is prose wrapped around JSON, and prose rots silently: a renamed component or a retired metric kind leaves the markdown reading exactly as before and the document beside it dead. So the mechanical half of the format's checklist is a test:

- every document loads and validates **with `--data`**, so a metric naming a dropped column fails here rather than after a GPU is booked;
- every workflow's inner documents resolve, `set` overrides applied;
- every relative link points at a file that exists, and every committed figure is actually shown (a figure carries no record, so an unreferenced one is a binary nobody can date);
- the seven sections are present and in order, the header table is complete, and `Reproduced` is one of its two legal forms;
- **every digest a demo quotes is one its own documents and tables really have.** This is what makes "pasted output, not typed by hand" enforceable — editing a `description` moves a digest, which is exactly the edit an author makes without thinking to re-paste. It caught three stale digests in this PR before it was opened.

Validation is pure (no weights, no network, no accelerator — spec §9), so this runs in the CPU tier CI already selects and costs under a second. Each check was verified to fail on a deliberately broken demo before being kept.

### 2. `causalab/protocol/registry.py` — register `meta-llama/Llama-3.2-1B-Instruct` (+19 lines)

The onboarding demos' model. `validate` and `explain` derive featurizer widths and layer bounds from this static table rather than fetching a config — that is what keeps them runnable offline and a digest independent of connectivity — so a document naming an unregistered key is refused by the pure verbs (`[V4]`) and is checkable only by running it. Without this entry the test above could not cover `02_trace` and `03_localize`, the two documents a new reader meets first.

The numbers are the checkpoint's own `config.json`, cited by revision in the comment beside them. Adding an entry moves no existing digest — `tests/protocol` is unchanged and green, including the pinned corpus.

The alternative was writing the onboarding demos on an already-registered model (gpt2, or Llama-3.1-8B-Instruct). Rejected: the reference figures are 1B, and an 8B model is the wrong size for the demo a newcomer runs first.

### 3. `README.md` — 3 hunks

The History section said the notebook demos were retired, which is now half the story: they return as `demos/`. Plus a Quick Start step pointing at the one demo that needs no GPU, and `demos/` in the repository-layout block.

`docs/running_experiments.md` §9 gains two rows in its "where to look next" table — inside `docs/`, listed here only for completeness.

## Verification

- `pytest -m "not golden"` over `tests/protocol tests/workflow tests/demos tests/causal tests/io tests/analysis tests/tasks` — **1736 passed**.
- `ruff check` / `ruff format --check` clean on the new and changed Python.
- Every demo document `validate`d and `explain`ed through the real CLI; every quoted digest is machine-checked by the new test.
- Demo 01's numbers were actually computed on CPU today, from the snippet the demo shows, which is why its header reads `✓`.
- The engine tests fail in my local venv on transformers 4.57 against this branch's `>=5.16` requirement. Unrelated to these changes: every failure and error in the full local run is under `tests/neural/engines`, and CI installs with `uv sync --frozen`.


can-goodfire:
## GPU evidence: all three documents run, on one H100 80GB

Pushed as bcfa16d. Ran on cw-rno2 (SLURM `1434777` / `1434785` / `1434787`), reference engine `pytorch_hooks`, bf16, models from the shared HF cache offline. The documents are byte-identical between `c26d7d9` (what ran) and `cf8a742` (branch head) — only markdown and tests moved between them — and all three `validate` digests reproduced on the cluster before the runs: `0685b0fa…`, `7dd122b2…`, `8143a336…`.

| demo | wall clock | header now |
|---|---|---|
| `02_trace` | **49 s** | ✓ digest `0685b0fa…` |
| `03_localize` | **39 s** | ✓ digest `7dd122b2…` |
| `weekdays_geometry` | **132 s** | ⚠ — numbers all regenerated, 4 figures still reference |

The estimates in each `Hardware` note ("not a measured wall clock") are now measurements.

### Three things the run found that reading could not

**1. Neither workflow demo's pasted `run` command works.** `causalab run <workflow> --dtype bf16` is refused: *"--dtype sets model.dtype on one protocol document; a workflow's steps each declare their own realization."* Both `03_localize.md` and `weekdays_geometry.md` printed it. Every protocol already pins `"dtype": "bf16"`, so the fix is dropping the flag — done, plus a callout in each demo and in `demos/README.md`.

**2. 02's hop is between L13 and L14, and no layer carries both columns.** The reference figure had one row (L12) where both the symbol and the slot were legible, and the prose read the hop off that overlap. On the document's own pair there is no such row: column 16 is 1 for L0–L13, column 21 is 1 for L14–L15, and the twenty other columns are 0 at every depth. Clean partition, two layers later, no overlap band.

**3. `weekdays` does not run at the layer anyone thought — it runs at L26.** `best` is a `select` step, so the layer is `locate`'s argmax. The prose said 18, the old notebook config said 28; the data says **26** (0.953, on an L26–L31 plateau). The workflow still *declares* `best_layer: 18` in `best.outputs.values.keys`, which is what the plan-time **authored** digests for `harvest`/`fit`/`walk` are computed against — so `explain` describes a plan at 18 while the run happens at 26. Recorded in Limits; reconciling it moves the workflow digest, so I left it as a deliberate follow-up rather than changing your experiment.

### What it confirmed

- **03's Q4 ceilings are exact.** symbol1 at `-6` hits **0.531** — 34/64, to the bit — for nine consecutive layers; symbol0 at `-10` hits 0.453 (29/64) against its 0.469 ceiling. So both symbol columns carry the variable *perfectly on the rows they apply to*, and the entire deficit from 1.0 is the position spec's, exactly as the document derived before running.
- **03's Q2 agrees with 02 to the layer:** `best/values.json` is `{"best_layer": 14, "best_pos": {"index": -1}}`, and 02's own run flips the slot at L14–L15.
- **weekdays' L18 handoff survives the change of counterfactual design:** 0.281 at L17, **0.891 at L18**.
- **The entity column sits at the floor across all 32 layers** (0.141–0.219, mean 0.181) — precisely what this document predicted independent resampling would cost, against the reference's 0.98 under entity-only resampling. And 0.219 is not *approximately* the floor: it is exactly the value at essentially every unresponsive cell, so the readable range of that grid is 0.219–1.0.

### Two "no result" sections now have one

- **02 Q4** — the distractor column (`index 12`) is **0.0 at every one of 16 layers**. The reference couldn't answer this; its counterfactual changed only one symbol.
- **weekdays RQ3b** — the DAS sweep ran. `k=2` → **0.641** (0.656/0.672/0.594), `k=8` → **0.984** (all three seeds), `k=32` → **1.000**. So two dimensions are enough to *draw* a ring and not enough to *be* the variable, and an 8-dimensional rotation beats intervening on the raw 4096-dimensional residual (0.953). New figure `rq3b_iia_by_k.png`.

### One verdict withdrawn

**RQ4's "a straight line crosses over" does not reproduce under this document's construction.** The reference walked between class centroids in a 32-d PCA subspace and found more non-weekday mass than any weekday at the midpoint. `weekdays_linear_walk.json` walks between two **rows'** activations in the **full** space at L26: only **9.5%** of interior points (55/576) put non-weekday mass on top, and on a typical row it *declines* monotonically. What survives is weaker and better supported — the line passes *through* a third weekday on **9 of 64** rows. The ring-vs-line comparison stays unanswered because the geodesic arm needs a script step that doesn't ship.

Also: aggregating that table across rows is the wrong move, and the demo now says why — each row interpolates between its own two answers, so the mean smears 11 endpoint pairs into a curve whose argmax is "Sunday" at every α. And 15 of 64 rows interpolate between two activations with the *same* answer day, which is the generator's failure to deconfound showing up a third time.

### Two smaller things

- **02 now ships no figure**, deliberately. A bare `causalab run` of a protocol writes `protocol.json`, not a step record, so `axes_for` returns `()` and `causalab.io.plots.workflow_figures` collapses all 352 cells into one. The pasted 16×22 table is the better artifact anyway — but a bare-protocol demo that wants a picture currently has to become a one-step workflow. Recorded in that demo's Limits.
- **`fit_pca`'s spectrum is normalized over the 32 components it keeps** (they sum to 0.9959), not over the full 4096-d space. So RQ3a can't state what fraction of full variance the subspace retains — the reference's 98% was that quantity and this run's 99.6% is not. Noted in Limits; a full-space denominator needs `fit_pca` to write the total variance alongside.

`weekdays` stays ⚠ because RQ3's 2D scatter / 3D manifold and RQ4's two walks are still pre-refactor. `Next` now names the three script steps that would close it — a 2D PCA scatter, a spline fit through the centroids (also RQ4's missing arm), and a multi-series probability-vs-α plot. All three are `script` steps over artifacts this run already produced (`pca/basis.safetensors`, `harvest/acts.safetensors`, `walk/day_probs.json`), so none needs a new document or another GPU hour.

Verified before pushing: `pytest tests/protocol tests/workflow tests/demos -m "not golden"` → **652 passed**; `ruff check` / `ruff format --check` clean. Every digest quoted in the markdown is still machine-checked by `test_quoted_digests_are_current`.

can-goodfire:
## Follow-up: digest reconciled, 02 has a figure, and a correction to my last comment

Three commits since the evidence drop: `41bbeb4`, `56685e4`, `757a417`.

### Correction — I got the authored-digest mechanism wrong

My previous comment said the declared `best_layer` in `best.outputs.values.keys` "is what the plan-time authored digests for `harvest`, `fit` and `walk` are computed against." **That is not true**, and changing 18 → 26 proved it: those three digests are unchanged (`4548aa31…`, `386b7941…`, `95cadecd…`).

What the declared value actually does is get substituted into the downstream documents at *load* time, where it is bounds-checked. Declaring 999 is refused:

```
refused: [W8] at steps.harvest document '../protocols/weekdays_harvest.json' does not load:
  [V4] at sites.target.layer site 'target': layer 999 out of range for the 32-layer model 'meta-llama/Llama-3.1-8B'
```

So it is a **validation placeholder**, and the authored digests are the step's document with its `set` block applied *symbolically* — which is exactly what makes them "authored" rather than "campaign" digests, the artifact's value being a runtime fact. That is a better design than the one I described. Corrected in the demo's Limits and in `Next`.

### The digest is moved: `8143a336…` → `b4c84874…`

`best_layer` now declares **26**, the layer `locate` actually selects. Re-pasted the inlined verbatim copy, the `validate` and `explain` blocks, the header and the Results provenance.

**Confirmed inert.** I re-ran the whole workflow under the new digest into a fresh output directory with nothing resumed (job 1435026, 117 s), and diffed it against the original:

| | declared 18 | declared 26 | same |
|---|---|---|---|
| `best/values.json` | `{"best_layer": 26, "best_pos": {"index": -1}}` | identical | ✓ |
| baseline accuracy | 0.828125 | 0.828125 | ✓ |
| locate max IIA | 0.953125 | 0.953125 | ✓ |
| PCA PC1 ratio | 0.224645365 | 0.224645365 | ✓ |
| PCA cumulative(32) | 0.995932545 | 0.995932545 | ✓ |
| `iia_by_k` k=2 / 8 / 32 | 0.640625 / 0.984375 / 1.0 | identical | ✓ |
| walk records | 704 | 704 | ✓ |

Bit-identical throughout. The declared value is validated against, not measured with.

### 02 has a real figure, via the one-step workflow

You're right that a one-step protocol with a plot after it is a perfectly good workflow — so I stopped treating it as a limitation and wrote it. `demos/onboarding_tutorial/workflows/mcqa_trace.json` is one `intervention_protocol` step naming the existing `mcqa_trace_scan.json`, plus one `heatmap` script step. Ran in **36 s**, and the figure is the same 352-point grid the Results table shows — index 16 bright across L0–L13, index 21 across L14–L15, no overlap row.

The scan step's campaign digest is still `0685b0fafc6037ae…`, so the figure was added without touching the experiment.

New section at the end of `02_trace.md`, *The one-step workflow* — about the shape rather than the result:
- why a protocol cannot draw its own table (`axes_for` reads the step record only a workflow writes, so a bare protocol run gives `KeyError: 'sites.target.layer'`);
- why the scan step *names* the protocol instead of restating it, and that the campaign digest proves it;
- why `plotted` belongs beside every `figure` — a `.png` carries no record;
- and when to reach for which: a protocol if the product is a number, a workflow if it is a figure or a value a later step consumes.

`demos/README.md` gains the same choice as two sentences under *Adding one*, so the next person adding a demo doesn't rediscover it.

### One more caveat the run surfaced

All seven weekday answers are tokenizer-ambiguous — `" Thursday"` is token 7950 space-prefixed and 38888 bare, both single tokens naming different rows — and `token_form` is left on `auto`, which takes the space-prefixed form for `logit_diff`, `cross_entropy` and the class probabilities alike. Correct for this template, since the answer follows `"is "`; wrong on a template whose answer starts a line, and unstated by the documents either way. Added to weekdays' Limits.

Verified: `pytest tests/protocol tests/workflow tests/demos -m "not golden"` → **656 passed** (61 in `tests/demos`, up from 57 — the new workflow document is covered by the validate / inlined-verbatim / linked / digest checks), `ruff check` and `ruff format --check` clean.

PR #76, opened 2026-08-31
The library defects the A3B protocol runs found

can-goodfire:
Three agents ran the six-step causal protocol end to end on `Qwen/Qwen3.6-35B-A3B`, one research question each — ~2 h of H100 time, 21 SLURM jobs. They found nine library defects, several of them independently. This fixes the ones whose fix is bounded, adds the tests that would have caught them, and ships the presets and APIs the runs had to hand-roll.

The audit was performed on `can/a3b-engine-parity`, which PR #74 has since merged into this branch — so this is an audit **of PR #20**, not of something stacked on it. Base: `can/protocol-refactor` @ `b6fc9aa`.

Plan: `notes/causalab-pr20-audit-fixes-plan.md` · findings: `notes/causalab-six-step-skill-audit.md` · triage: `notes/causalab-six-step-audit-remediation-strategy.md`. The skill-tree half is a separate silico PR; three of its paragraphs describe things built here.

## The headline: a two-word class of bug that broke §1.1's own promise

A split method/application document raised an unhandled `TypeError: unhashable type: 'dict'` whenever the application swept `sites.<name>.component`, while **the same experiment written flat loaded fine** — contradicting the spec's own promise that a split document "digests exactly as the same experiment written flat". `method.py:302` tested membership in a frozenset of strings with a `{"sweep": [...]}` dict on the left; its twin in `canonical.py` already guarded with `isinstance(component, str)`.

Flat and split now digest identically (`8644b9b7…` on the fixture pair), which is the regression test and the §1.1 property test in one assertion.

**This retracts audit finding c3** — "`sites.component` is not sweepable, so `attention_output` and `mlp_output` are two documents and two 35 B loads." It *is* sweepable in a flat document; the failure was always this `TypeError` in the split path.

## The fit path: step 4's "report train and eval together" becomes satisfiable

- **`train.eval` reached nothing.** `_run_eval`'s return value was consumed only inside the `early_stop` branch and dropped, so a fit document's `iia.json` was the **train** score under a name every reader took for the eval one. It cost the refusal run five extra apply documents and five extra 35 B loads to work around. Now a sibling `train_eval.json` — a sibling, not a metric column, because it is a different split and so a different population.
- **`early_stop` never restored the best weights.** `best` was tracked; nothing snapshotted the parameters. A run that early-stopped after `patience` non-improving evals saved the *worst of the tail*. The question-mark run saw held-out 1.000 at every seed, so last and best coincided at ceiling — luck, not a property.
- **A DBM fit now says whether its mask is a mask.** The shipped preset produced **0 of 2048** dimensions outside [0.1, 0.9] and still scored **1.000** at layer 38, because `Gate._mask` is a *hard* `θ > 0` mask in eval mode and an unseparated θ makes it a coin flip on gradient noise. `fit_diagnostics.json` records `decisive_fraction` and `hard_mask_size` per gate. **The retune is not done** — it needs a GPU run to validate; the preset carries the measurement and the warning in its own description instead of shipping quietly.

## Refusals that used to be wrong numbers

- **`token_form: "auto"` refuses instead of guessing.** It already *warned* when both forms were single tokens naming different rows — and the run read a flat 0.000 at all 48 layers of a real gpt2-xl scan anyway. The silent half (`" 85"` → `[220, "8", "5"]`) is two different problems, and both are now refusals: under `first_token`, a value whose two forms credit *different* first pieces; and an answer space that is not first-token distinct, where an emitted `87` scores 1.000 against an expected `85`.
- **Ragged writes refuse by rule 19, before the forward pass.** They used to raise a bare `NotImplementedError` while landing the write — on the accelerator, after 35 B of weights had loaded. It shaped a corpus: every request had to end in a `.` so negative indices aligned.
- **A double BOS is refused at encode.** With no chat-template path, both runs baked the rendered template into the dataset's `input`; `encode()` then added a second BOS and every position shifted by one, silently. §2.3 now also states that `prefix_lengths` is 0 in v1, so the spec's rebase rule is an identity — a rule with no implementation is worse than a missing one.

## What the runs had to hand-roll, now shipped

- **`--register-from-hf` on `validate`/`explain`/`digest`.** All three runs wrote the same nine-line wrapper. The documented workaround — validate against a similar registered model — gives a **false** refusal (`[V4] layer 36 out of range for the 36-layer model 'Qwen/Qwen3-4B-Instruct-2507'` on a valid 40-layer document). Opt-in, so without it a digest still never depends on the network; workflow-aware, because a workflow names several documents.
- **`choose: "knee"`** in `select.py` — step 4 says *choose rank from the IIA-versus-k curve, not the highest score*, which `max` cannot express. On the refusal run's saturated curve `idxmax` returned k=2 by table order: right by luck.
- **The matched-k random-subspace control**, required by the report format at every cell and provided by no preset. This needed one field the plan did not expect: an untrained `subspace` *is* a random orthonormal basis, but its seed came from `train.seed`, so a document with nothing to train could not vary the draw. `seed` is now an authored `subspace` field, defaulting to the document's seed — **no existing document's canonical form or digest moves.**
- **`serialize_examples`** — there was no public path from a hand-authored causal model to a serialized table, so step 3 dead-ended at "now build the task package" after telling you to write `models.py` in your own directory.
- **`explain --engine`** prints which engine would run, or the §8 refusal. Routing could not be pre-flighted at all, on the one model where `delta_*` is hooks-only and `deltanet_*` is nnsight-only.
- **Two examples the docs owed**: a multi-position patch (positions −4/−3/−2 read 0.000 individually at all 40 layers and 0.367 → 1.000 jointly — invisible to any single-position scan), and mean replacement as the two-document `reduce: mean` → `params` chain.

## Decisions to review

**`--engine` now defaults to `auto`** (the audit's explicit ask). `auto` is every installed engine with the reference **first**, and list order is preference, so anything the reference serves behaves exactly as the old `pytorch_hooks` default did. What changes is the other case: a document only the nnsight engine can serve runs instead of refusing by name. A document nothing serves still refuses (`ValidationError` 13).

**`token_form` pins move 32 documents' digests, and no values.** The refusal is tokenizer-dependent, so the CPU suite (tiny-random sentencepiece, where `" X"` and `"X"` are one id) cannot see it. Against the real tokenizers the documents declare, **95 metric columns in 32 documents** would refuse. Each is pinned to `space_prefixed` — exactly the form `auto` was already choosing — verified over **1283 (value, column) resolutions with 0 mismatches** by resolving every pinned column both ways against the cached tokenizers. Digests move (the pin is canonical form); `corpus_digests.json`, `workflow_digests.json` and `golden_digests.json` are regenerated. `drift_goldens.json` and `paper_goldens.json` key by metric name, not digest, and the token ids behind them are unchanged. `hours_baseline_im.json` stays on `auto` — its answers are `"00".."23"`, which follow no space.

**Two deviations from the plan, both because the plan's premise did not hold:**

1. The plan put the ragged-write width check in `validate --data`. It cannot go there: for an `all` or `variable` position the width is the *tokenized* length, and `validate` holds no tokenizer — `ResolutionEnv` carries datasets and artifacts only, and the pure verbs are deliberately torch- and network-free. Encoding is the earliest point the question has an answer, so rule 19 fires there, and §5 says so where the rule is defined.
2. The plan asked `first_token` to refuse *any* multi-token value under `auto`. That was implemented, measured, and reverted: it deletes the designed prefix-grading path (`" Thursday"` → `Th`) that the weekdays end-to-end pins rely on. The two narrower refusals catch the same failures without that cost.

## Out of scope, filed with names

`--max-batch` (not one flag — `executor_base.py:434` encodes a whole role into one `EncodedBatch`; its own PR with its own parity test, and the biggest item left), a first-class `chat` field, two sites sharing one sweep axis, `select` emitting a *set* of cells, `positions` in the method half, a multi-hypothesis panel primitive, per-digit answer forms, and `distinguishability_report`'s deepcopy cost.

**Retracted, do not encode anywhere:** the audit's `rsync`-hangs-on-the-login-node report does not reproduce — ~20 syncs of a 73 MB tree with no delay, and every `rsync -a` under a second.

## Gates

`uv run pre-commit run --all-files` clean; `uv run pytest -m "not golden"` → **2721 passed** (2669 on the base). The GPU golden tier is unrun locally: no golden *document* changed except by the `token_form` pin, and that pin is value-preserving by the verification above.

---

# Second batch — the `research/` pipeline audit

The five commits after `824c1d8` close a **different** audit from the one this
PR was opened for. This PR was written against the earlier *silico-skill* audit;
these close the **`research/`-pipeline** audit of the same six-step protocol as
shipped in PR #20 and exercised by three parallel A3B investigations. The two
overlap, and every item below was re-checked in code on this branch's tip rather
than inferred.

The structural reason they are here: **PR #76 as it stood changed 0 of the 63
files under `research/`**, and `research/` is in this tree (it comes from the
base), so the fixes land in the same PR.

## Status of all 14 items

| # | Item | Before this batch | Now |
|---|---|---|---|
| 1 | Qualify step 4's stopping rule to the readout cell | open | **closed** — `hypothesis-testing.md`, plus the same scoping in `residual-stream-das.md` and both DBM docs |
| 2 | Held-out numbers come from a separate *apply* document | library landed, doc open | **closed** — `iia.json` / `train_eval.json` / apply-document table in `residual-stream-das.md` and both DBM docs |
| 3 | Fix the `gate` featurizer load path | open | **closed** — `Gate.from_theta` + dispatch + `dbm_apply.json` + 7 tests |
| 4 | `weekdays_locate_scan` tap and `validate --data` | open | **closed** — and larger than reported; see below |
| 5 | Delete "one document per band" | open | **closed** — replaced by a worked `intervened_models` example in three documents, shipped as `attention_band_patch.json` |
| 6 | `--register-from-hf` | landed | landed |
| 7 | DAS criterion must reference null + ceiling | open | **closed** — `residual-stream-das.md`, plus `decisive_fraction` required beside every DBM effect |
| 8 | Make `token_form` required | 32 documents pinned, schema open | **closed** — required on every kind in `TOKEN_COLUMN_METRIC_KINDS` |
| 9 | Give step 1 a real execution recipe | open | **closed** — stub retired; `research/README.md`'s inventory drops to one |
| 10 | MoE/hybrid guidance in the method docs | open | **closed** — "Non-Llama towers" in both DBM docs |
| 11 | Grouped head gate / scope attention-head DBM | open | **doc half closed** (stated in the flow chart, not only the leaf); library half remains out of scope |
| 12 | Batch-size knob | out of scope | still out of scope, unchanged |
| 13 | Step 6's PDF | open | **closed** — `report.md` + self-contained `report.html`; there is no LaTeX on the cluster |
| 14 | Smaller items | mixed | **closed**: `model_dtype`, partial `entry`, base/cf cross, empty axis, chat template — see the last commit |

**14 of 14 addressed: 12 closed here, 1 already landed, 1 (item 12) deliberately
left as its own PR. Item 11 lands its doc half only, by the same reasoning.**

## Three places this audit's plan was wrong, and what was done instead

Recorded because each cost a verification cycle and each changes what the fix is.

1. **"No data in the repository has a `subject` column"** — false.
   `tests/protocol/fixtures/data/weekdays/{train,test}.json` both have one, as do
   the golden counterfact/drift fixtures. It is true only of *task-generated*
   tables, which is precisely why the fixture-backed tests never saw the defect.
2. **Retargeting the tap `subject` → `entity` is not enough.** Verified by
   running the shipped document: `[V19] … addresses ragged position widths
   [1, 3]`. A bare `{"variable": …}` window is ragged wherever the value
   tokenizes to different lengths, which weekday names do. The tap is now
   `{"index": -1, "scope": {"variable": "entity"}}` — the last token of the
   span, the idiom `tests/golden/protocols/rome_aie_*` already uses.
3. **The false green had a second, larger cause.** `check_data_columns` read
   `point_documents[0]`, so a swept axis was only ever checked at coordinate 0 —
   even the column check it *did* run could not have seen a bad coordinate in a
   sweep. It now checks every expanded point, with table reads cached per
   dataset ref.

## Decisions made rather than deferred

- **A partial authored `entry` is *not* completed from the consuming point's
  coordinates.** The plan allowed either that or an explicit refusal. Chosen:
  keep authored-wins, because completing one spelling from the other would make
  a selector's meaning depend on which axes the consumer happens to sweep. The
  behaviour was already loud rather than silent; what is new is a refusal that
  names the disambiguating coordinates and says pinning `k` elsewhere does not
  narrow an `entry` that omits `k`. **No resolution changes.**
- **The base/counterfactual *n*×*n* cross is documented, not refused.** There is
  no zip to redirect to — axis identity is name identity and `dataset` is
  required on every role — and the intended shape is one table carrying both
  sides, which every shipped preset uses. Refusing a shape with no alternative
  is worse than the hazard. This is the same "two fields sharing one axis" item
  already filed above as out of scope.
- **The chat-template machinery is documented, not deleted.**
  `apply_chat_template` is called nowhere; `token_positions.py` now says so at
  the top, and marks its `use_chat_template` field and chat-prefix hook as the
  Plan-era pipeline surface the task packages annotate against. Not deleted
  because they are live parameters of that legacy module, not unreachable code.

## Corrections to this PR's own first batch

- **`hours_baseline_im.json` no longer "stays on `auto`".** Making `token_form`
  required forced the question, and the answer is `bare`: the document's own
  description says the paper template ends in a trailing space "so the
  zero-padded hour is the single next token", so the answer is emitted with no
  space before it. `bare` is the form `auto` was already resolving to, so **no
  value moves** — but a reflexive `space_prefixed` would have named a token the
  model never emits. That is the argument for requiring the field, arriving on
  the first document to be asked.
- **`weekdays_das_apply.json` now declares `"dtype": "bf16"`.** It declared
  nothing, and a `model` block with no `dtype` implies fp32, while
  `weekdays_das_sweep.json` declares bf16 — so the shipped fit → apply pair, and
  `weekdays_8b.json` which chains them, could not run at real scale. Verified
  through `canonical_model` on both presets.

## Digests

| file | moved by | why no value moved |
|---|---|---|
| `corpus_digests.json` | the weekdays fixture rename (`subject` → `entity` plus a `counterfactual_inputs_variables` sibling) reaches every document reading `weekdays/*` through the dataset content digest; 01, 07 and 09 also change themselves | the locate scan's 16 tiny-scale cells score identically at the shared `{"index": -1}` half, and its other half previously could not run at all; 09's pair could not run at its declared realizations |
| `workflow_digests.json` | follows the documents its steps name | as above |
| `golden_digests.json` | `hours_baseline_im.json` alone, for its `token_form` pin | the pinned form is the one `auto` already resolved |

## Gates

`uv run pre-commit run --all-files` clean. `uv run pytest -m "not golden"` →
**2740 passed** (2721 at `824c1d8`); the full `uv run pytest` is 2740 passed,
27 skipped — the skips are the GPU golden tier, which **cannot be run on the
Mac and was not run**. All 109 relative links in `research/` resolve.

---

## Merge of `can/protocol-refactor` (`9d814ae`)

The base moved 13 commits while this batch was in flight — PR #75
(demos-as-markdown) and the backend-forward-interning merge. Merged in rather
than rebased: 17 commits over three regenerated digest files is not a rebase
anyone should review.

**One textual conflict**, in `neural/shared/execution.py`'s `__all__`: both
sides added names, both stay.

**One semantic conflict git could not see, and it is the interesting half.**
This batch makes `token_form` required on the token-column metric kinds; the
base added six demo protocol documents written before that rule. Nine of their
metrics stopped parsing, every workflow naming them stopped loading — 11
failures on the merged tree, and a fair test of whether the requirement is
worth its cost.

It is, and the demo said so itself before I got there. `weekdays_geometry.md`'s
Limits already carried:

> `" Thursday"` is token 7950 space-prefixed and 38888 bare, and both forms are
> single tokens naming different rows, so `auto` picks the space-prefixed one …
> Setting each metric's `token_form` to `space_prefixed` explicitly would make
> the demo say what it relies on.

So every pin is `space_prefixed`, every pin is the form `auto` already resolved,
and **no number and no figure changes**. That Limits bullet now records the
choice the run made implicitly instead of listing it as a gap. The MCQA demos
are the same case (`" Z"`/`" Q"` after `Answer:`).

Three checks the demo format enforces, all re-derived rather than hand-edited:

- **inlined copies** — every changed document's verbatim block re-pasted from
  the file's bytes.
- **quoted digests** — 609 of the demos' 623 digests moved. Rather than match by
  eye, every digest was dumped from a throwaway worktree at the base and again
  here, keyed by (document, kind, index), and each quoted prefix replaced with
  *its own* new value. 24 substitutions, no ambiguous prefixes.
- **`Reproduced` headers** — these now name the digest the run actually had
  **and** the digest the document now has, with the reason they differ. Quoting
  only the new one would have claimed a run at bytes nothing has executed. The
  one narrative before→after digest pair in weekdays_geometry's Limits got the
  same repair: a later edit had overtaken its "after".

`uv run pytest` → **2807 passed**, 27 skipped; `pre-commit run --all-files`
clean.


can-goodfire:
## The red `test` gate is inherited from the base, not from this PR

`test_generate_frame.py::test_a_decode_moves_prompt_frame_reads_only_by_float_noise` fails here. It is **not this branch**, and it will not clear on a re-run:

- **The base's own head fails it.** `can/protocol-refactor` @ `b6fc9aa` — the commit this PR is based on — fails the same test on its own CI ([run 33358044130](https://github.com/goodfire-ai/causalab/actions/runs/33358044130)). `fb0dbb22`, the commit before it, passed ([run 33356765087](https://github.com/goodfire-ai/causalab/actions/runs/33356765087)).
- **`b6fc9aa` is markdown-only** — `research/**` docs, no Python at all. So no code change caused the flip; CI's environment did.
- **It is no longer intermittent.** It was seen twice on 2026-08-28 and passed on re-run; here it failed twice consecutively. It passes 3/3 on the Mac.
- **This branch shares one file with the test's subject**, `neural/shared/encoding.py`, where the change is a read-only guard plus a docstring — it raises or returns and cannot move a tensor.

### What is actually wrong with the test

The behavioural claim is the assertion above the failing one — *a decode moves a prompt-frame read by no more than `atol=1e-7`* — and it passes. The failing line is a **premise guard**:

```python
assert not torch.equal(plain, also), (
    "if these became bit-identical, use_cache stopped changing the kernel "
    "path and this test's premise (and its tolerance) should be revisited"
)
```

It asserts an implementation detail — that `use_cache=True` takes a different kernel path and therefore accumulates *different* float noise. That is not a property of the code, so it is free to become false, and on CI's CPU it now has. Bit-identity is a **stronger** result than the test demands, being reported as a failure.

The fix is to drop that guard and keep the `allclose` assertion exactly as strict — not to widen the tolerance, which would weaken the real claim. **It belongs on `can/protocol-refactor`**, where the red starts, rather than in this PR, so that the base stops shipping a red gate to everything based on it. Happy to send that one-line PR if you want it.

PR #77, opened 2026-08-31
Split Jupyter-only deps into a notebook extra (port of internal #539)

can-goodfire:
Re-port of [causalab-internal #539](https://github.com/goodfire-ai/causalab-internal/pull/539) (@ryan-goodfire) onto the intervention-protocol tree, following the [public→internal sync](https://github.com/goodfire-ai/causalab-internal/pull/545) that superseded it. The pyvene half of #539 is moot — pyvene is no longer a dependency.

**What**: `ipywidgets`, `jupyterlab`, `ipycytoscape`, `jupyter-dash`, `ipykernel` move from core deps to a `notebook` extra. `dash`/`dash-cytoscape` stay core (`causalab.io.plots.causal_graph` imports dash at module level).

**Why**: the original motivation stands on the new tree — headless installs (the silico cluster-job GPU image merges causalab into a network-less job venv) shouldn't carry the jupyter-server stack.

**Verified**:
- `grep` over `causalab/ tests/ scripts/ research/`: zero imports of the five packages (module-level or lazy)
- `uv export --frozen`: 0/5 in the default dependency set, 5/5 with `--extra notebook`
- `uv lock --check` passes → CI's `uv sync --frozen` unaffected

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,9 +15,6 @@ dependencies = [
     "pandas",
     "numpy",
     "scikit-learn",
-    "ipywidgets",
-    "jupyterlab",
-    "ipycytoscape",
     "seaborn",
     # 5.16 is the first release carrying models/qwen3_5_moe (Qwen3.6-35B-A3B's
     # text tower); the hookpoint vocabulary work targets that architecture.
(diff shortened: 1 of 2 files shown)


can-goodfire:
Wrong target: this port belongs in causalab-internal, stacked on the public→private sync PR (goodfire-ai/causalab-internal#545), so the superseded internal work re-surfaces there. Re-opened as a causalab-internal PR based on can/sync-public-main.

PR #78, opened 2026-08-31
The seven demos main still had as notebooks

can-goodfire:
Finishes what #75 started. `main` had eleven onboarding notebooks; #75 translated four (`01`, `02`, `03`, `weekdays_geometry`). These are the other seven, in the `docs/demos.md` format, each with GPU evidence gathered the way [#75's evidence comment](https://github.com/goodfire-ai/causalab/pull/75#issuecomment-5482479993) did it.

| notebook on `main` | demo | question | answer |
|---|---|---|---|
| `causal_model_demo` | [`causal_model/`](demos/causal_model/causal_model.md) | two algorithms agree on every input — does an intervention tell them apart? | yes, **914/1000** against an exact 0.900 floor. The *output* tells them apart on **0**, by identity |
| `04_train_DAS_and_DBM` | [`04_subspace`](demos/onboarding_tutorial/04_subspace.md) | how few directions carry the answer symbol? | **32–64 of 2048**. k = 1 is 0.000; k = 64 matches the whole stream |
| `05_PCA_DBM_DAS` | [`05_variance_vs_cause`](demos/onboarding_tutorial/05_variance_vs_cause.md) | is the top-variance direction the causal one? | **no** — PC0 is 7.65 % of the variance and **0.000** IIA |
| `06_boundless_DAS` | [`06_components`](demos/onboarding_tutorial/06_components.md) | attention or MLP? | **attention, at one layer** — 0.719 at L14, every MLP ≤ 0.016 |
| `07_cross_model_patch` | [`07_cross_model`](demos/onboarding_tutorial/07_cross_model.md) | can one checkpoint's activation drive another? | **it loads and is refused**; the same graft inside one checkpoint reproduces 03 at 0.969 |
| `08_attention_pattern_analysis` | [`08_attention`](demos/onboarding_tutorial/08_attention.md) | pattern or value, and which head? | **the value, at L14 H11** — 0.375, with 125 of 128 cells exactly 0.000 |
| `10_steering_interventions` | [`10_steering`](demos/onboarding_tutorial/10_steering.md) | which layers are necessary, can one direction suffice? | **15 of 16 necessary**; the mean-difference direction flips **1 pair in 64** |

Every `Reproduced` header is `✓` with a date, an engine and a digest. No figure and no number here predates the document above it.

## The through-line

`04`–`10` all sit at **the cell `03` located** — block 14's output at the answer slot, IIA 0.969 — and vary one thing each: the subspace (04, 05), the site (06, 08), the model (07), the mechanism (10). Read in order they are one cell from six directions, and they agree: the variable is *wide* (32–64 directions, 1029+ of 2048 gate dimensions), *late* (one layer), and *delivered rather than routed* (values, not patterns).

**Four of seven land on a negative or a refusal**, and each is a result rather than a gap. Where a null is a mathematical identity, the demo says so in the sentence that makes the claim — `causal_model`'s Q4 and `10`'s α = 0 control both are.

Two agreements neither scan could see coming:

- **`08`'s `select` step emits `{"best_layer": 14, "best_head": 11}`** — the head the notebook it replaces fixed by hand in its first cell and then characterized statistically without ever intervening on. The descriptive method and the causal one name the same head; the causal one adds that it is load-bearing on 24 of 64 pairs.
- **`07`'s in-model graft returns 0.9688** against `03`'s 0.969 — a value that made a round trip through a `.safetensors` file and a `params` entry and came back the same intervention.

## Decisions worth reviewing

**1:1 with the notebooks, keeping their numbers.** Seven notebooks, seven demos, `04`–`08` and `10` numbered as they were. There is no `09` because `main` had none. The alternative — merging 04/05/06, which all touch DAS and DBM — was rejected: each notebook already has a distinct question, and the mapping stays reviewable.

**A learned method ships a fit *and* an apply, and this is now in the format.** A fit document's own `iia.json` is the score on the split that chose it. `04` measures **0.945 train against 0.531 held-out at k = 16** — nearly a factor of two, worst exactly where a reader stops. `docs/demos.md` gains **§4.1** saying so, and `04`, `05` and `06` all take the shape. This is the durable half of what #76 made reachable for `gate`.

**Registered `meta-llama/Llama-3.2-1B`** (+19 lines in `causalab/protocol/registry.py`), the pretrained sibling of the demo model. `07` names both checkpoints and the pure verbs size a document from this table rather than fetching a config, so the second entry is what lets that document be checked offline. Values are the checkpoint's own `config.json`, cited by revision beside them, and verified against the cluster's cached snapshot. Adding an entry moves no existing digest.

**Two new tables** — `mcqa/train_n128_s1` and `mcqa/test_n64_s2`, seeds 1 and 2. `pairs_n64_s0` is already spent as the scan set `02` and `03` use, and a fit and its apply must not share rows.

**`tests/demos/test_demos.py` gains one narrow exception** (and only that). Every document under `demos/` is loaded standalone by `test_validates`; an apply document names its artifact by a run-tree path whose leading segment is a step name, which standalone is `[V15] artifact file not found` — the loader being right. The exception fires only on a V15 whose path does not exist under the repo **and** only when a workflow in the same demo names the document as a step, so coverage moves to that workflow's entry, which is stricter (it applies the step's `set` and checks the producing step really writes the file). #75 had no apply document, so nothing had met this.

**The six workflow demos follow §3's shape**, which `weekdays_geometry` sets and these did not at first: the workflow unfolded, then a step table built from its own `steps` (script steps included), then each step document folded into a `<details>`. A file that several steps run is inlined once and its summary names every step — 06's three gate fits differ only in an L1 weight applied by `set`. 07 gets a two-row table with a column per arm instead, because its two workflows declare the same step names running the same files.

**Where the demos are narrower than the notebook, the reason is measured and stated.** `08`'s band is layers 12–15 rather than all 16, for two reasons both in its Limits: `06` reads `attention_output` at 0.000 everywhere but L14, and the full grid does not fit (below).

## Three defects the runs found, none fixed here

These are recorded in the demos' `Limits` with their evidence, and left for a follow-up — they are engine work, not demo work.

**1. `sites.<name>.head` does not reach an `attention_probs` write.** The first version of `08`'s pattern arm swept 4 layers × 32 heads. Per-example `logit_diff` came back **bit-identical across all 32 heads** of a layer — `max |H0 − H31|` over 64 examples is exactly `0.0` at L14 — while differing across layers (`max |L14H0 − L13H0|` = 1.4375). The write replaces the whole layer's pattern. The document validated, the sweep expanded to 128 points with 128 distinct point digests, and 32 of every 32 were the same experiment. **A wrong result the record itself cannot catch**, which is why the arm now sweeps layers only.

**2. Write legality is checked at the plan, not at load.** `attention_result` is derived and unwritable, and the refusal says so well — it even names `attention_premix` as the alternative. But `causalab validate` and `causalab explain` both pass on the document that can never run. `validate --data` catches a metric naming a dead column before a GPU is booked; this class still costs a job, and it cost one here.

**3. A workflow step cannot be sharded, so a large grid has no fallback.** 16 layers × 32 heads × 64 pairs is 32 768 row-forwards planned as one campaign, and the executor batches points: `torch.OutOfMemoryError` in llama's own attention, **79.16 GiB in use of 79.18** on an H100 80GB — for a 1.2 B model. `--points START:STOP` shards a bare protocol run; there is no equivalent for a step inside a workflow.

## GPU evidence

Four SLURM jobs on `cw-rno2`, reference engine `pytorch_hooks`, bf16, models from the shared HF cache offline. Every demo's numbers come from the branch head's own documents.

| demo | job | wall clock |
|---|---|---|
| `04_subspace` | 1435244 | 88 s |
| `05_variance_vs_cause` | 1435244 | 24 s |
| `06_components` | 1435244 | 35 s |
| `07_cross_model` (control) | 1435244 | 17 s |
| `07_cross_model` (refused arm) | 1435244 | 16 s, stops after its harvest |
| `08_attention` | 1435273 | 48 s |
| `10_steering` | 1435244 | 22 s |
| `causal_model` | — | 0.40 s, CPU, best of three |

Job `1435241` was the first pass; three of its arms changed the documents rather than the prose, and [that commit](https://github.com/goodfire-ai/causalab/commit/7db96d0) records what and why — a k sweep that stopped short of its own answer, a mask that was decisive but not sparse, and the head grid that would not fit.

## Verification

- `pytest -m "not golden"` over `tests/protocol tests/workflow tests/demos tests/causal tests/io tests/analysis tests/tasks` — **1916 passed**.
- `ruff check` / `ruff format --check` clean; `pre-commit run --all-files` passes.
- Every quoted digest is machine-checked by `test_quoted_digests_are_current`; every inlined document is held to its file byte for byte by `test_is_inlined_verbatim`.
- Every demo document `validate`d and `explain`ed through the real CLI, and every `explain` block in the markdown is pasted output.


can-goodfire:
## GPU evidence: all seven demos, on one H100 80GB

Pushed as `2943405`. Ran on `cw-rno2` (SLURM `1435241` / `1435244` / `1435253` / `1435261` / `1435273`), reference engine `pytorch_hooks`, bf16, models from the shared HF cache offline. `1435244` and `1435273` are the evidence runs; the others are the iterations that got there.

| demo | wall clock | header |
|---|---|---|
| `causal_model` | **0.40 s** (CPU) | ✓ |
| `04_subspace` | **88 s** | ✓ digest `d1f96dd3213ba5b2…` |
| `05_variance_vs_cause` | **24 s** | ✓ digest `c7cd0e0bb5a4e0e4…` |
| `06_components` | **35 s** | ✓ digest `bb0a8ec10cd38eda…` |
| `07_cross_model` | **17 s** control, **16 s** refused arm | ✓ digests `ea55f32b723f356f…` / `6c1dcd04df3ea1a0…` |
| `08_attention` | **48 s** | ✓ digest `d191113a1388aa4b…` |
| `10_steering` | **22 s** | ✓ digest `a7955cba25723b42…` |

**All 60 numeric claims in the seven demos were re-derived from the run tree and matched** — every IIA, every train/test pair, every `hard_mask_size` and `decisive_fraction`, every ablation cell and every steering strength.

### Four things the runs found that reading could not

**1. `sites.<name>.head` is silently ignored on an `attention_probs` write.** The first version of 08's pattern arm swept 4 layers × 32 heads. Per-example `logit_diff` came back **bit-identical across all 32 heads** of a layer — `max |H0 − H31|` over 64 examples is exactly `0.0` at L14 — while `max |L14H0 − L13H0|` is `1.4375`. The write replaces the whole layer's pattern. The document validated, the sweep expanded to **128 points with 128 distinct point digests**, and 32 of every 32 were the same experiment. This is the dangerous class: a wrong result the record itself cannot catch. The arm now sweeps layers only, and the second arm uses `attention_premix`, where naming a head does select one head.

**2. DAS needs an order of magnitude more width than "a direction".** The first sweep stopped at k = 16 and was flat at zero — 0.000 held-out at every k but 16, where it was 0.047, while an *untrained* 32-dimensional PCA basis at the same cell reached 0.828. A curve flat at zero cannot distinguish "no such subspace" from "the fit did not reach it", so the sweep now runs to 128 at 20 epochs. It leaves the floor between k = 8 and 16, and **k = 64 matches the whole 2048-wide stream to within 0.016**.

**3. The DBM mask was decisive but not sparse.** `fit_diagnostics.json` reported `decisive_fraction` 0.972 — the gate committed — and `hard_mask_size` **1987 of 2048**. A held-out 1.000 from a mask that drops 61 dimensions says nothing about how many the variable needs. Three L1 weights later: λ = 3.0 gets to 1029 of 2048 and costs 0.17 of IIA, with `decisive_fraction` falling to 0.669. **No sparse decisive mask exists in that range**, which is the honest answer and the opposite of what the notebook's framing expects.

**4. The full head grid does not fit, and a workflow step cannot be sharded.** 16 layers × 32 heads × 64 pairs is 32 768 row-forwards planned as one campaign, and the executor batches points: `torch.OutOfMemoryError` in llama's own attention, **79.16 GiB in use of 79.18** — for a 1.2 B model. `--points START:STOP` shards a bare protocol run; there is no equivalent for a step inside a workflow.

### Two agreements neither scan could see coming

- **08's `select` emits `{"best_layer": 14, "best_head": 11}`** — the head the notebook it replaces fixes by hand in its first cell (`TARGET_LAYER = 14`, `TARGET_HEAD = 11`) and then characterizes statistically without ever intervening on. The descriptive and causal methods name the same head; the causal one adds that it is load-bearing on 24 of 64 pairs, and that **125 of 128 cells are exactly 0.000**.
- **07's in-model graft returns 0.9688** against 03's 0.969 — a value that made a round trip through a `.safetensors` file and a `params` entry and came back the same intervention. The two pairs it does not flip say a *third* thing: `" D"` at 0.377 and `" yellow"` at 0.399.

### Two refusals that are results

**`07`'s cross-checkpoint graft is refused**, before a forward runs:

```
[V15] params entry 'v_src' (source/acts.safetensors): ArtifactIdentity mismatch
on 'model_key' — the document implies 'meta-llama/Llama-3.2-1B-Instruct' but the
bundle was stamped 'meta-llama/Llama-3.2-1B' (§2.5)
```

Right, and worth a hard refusal rather than a warning: the shapes match (2048 and 2048), so the graft *would* have run and produced a plausible small number that nobody could distinguish from an honest one.

**`08`'s `attention_result` write is refused at the plan**, with the message naming its own alternative (`attention_premix`, same `head`). Both refusals arrive **after** `validate` and `explain` pass — write legality and run-tree `ArtifactIdentity` are checked when the campaign is planned, not when the document loads. `validate --data` catches a metric naming a dead column before a GPU is booked; this class still costs a job, and it cost two here.

### What it confirmed

- **06 and 08 agree to the layer and then narrow.** `attention_output` is 0.000 at fifteen of sixteen layers and **0.719 at L14**; every MLP in the model is ≤ 0.016. One layer down, `attention_premix` at (L14, H11) is 0.375 — one head of thirty-two recovering **52%** of what all thirty-two do together.
- **05's negative is sharp.** PC0 explains **7.65%** of the variance at the located cell and scores IIA **0.000**; `logit_diff` −8.54, the un-intervened margin. The 32-component basis reaches 0.828. Variance rank does not predict causal usefulness, and at the top of the list it is not even correlated with it.
- **10's steering negative is a property of the contrast, not the cell.** The mean base→counterfactual difference flips **1 pair in 64** at its best, while the same cell flips 62 in 64 under an interchange. `different_symbol` resamples the symbol per row from 26 letters, so 128 differences point in 128 unrelated directions and their mean is noise. The accuracy and `flipped` curves never cross — the signature of a norm effect rather than a semantic one.
- **10 also found something it cannot explain**: zeroing the answer slot's residual is fatal at fifteen of sixteen layers (eleven at exactly 0.000, against a clean 0.891) — but **L12 retains 0.469**, with both neighbours at 0.000. Recorded as unexplained, with the next document named.

Verified before pushing: `pytest -m "not golden"` over `tests/protocol tests/workflow tests/demos tests/causal tests/io tests/analysis tests/tasks` → **1916 passed**; `ruff check` / `ruff format --check` clean. Every digest quoted in the markdown is machine-checked by `test_quoted_digests_are_current`, and every inlined document by `test_is_inlined_verbatim`.

PR #80, opened 2026-09-01
Refuse a write operand read from deeper than it lands (§5.20)

can-goodfire:
## Why

Nothing checked that a write's operand came from an address the network can
route to it. A document could read a head's contribution at layer 20, add it
into the residual stream at layer 11, and get a number — one attributable to no
path, since no edge carries information backwards through the forward pass.

The reason it was never caught: the value is *computable*. §2.9 stages the
operand's model first, which is exactly what acyclicity licenses. So this is
**uninterpretable, not unexecutable**, and the loader only refused the latter.

This is the one insight worth keeping from the pre-protocol path-patching engine
in `causalab-internal#534`, whose `check_sender_reachability` refused nonphysical
sender→receiver edges. That engine's other three contributions this tree already
has independently — joint multi-edge routing (unbounded additive writes in one
intervened model), the q/k/v head-space receivers with GQA-narrowed head bounds,
and a derived `attn_eager` requirement — but nothing had replaced the
reachability check.

## What

Rule 20: every write operand naming a **read** is read at an address no deeper
than the one the write lands on, in the `(layer, intra-block rank)` order of
§2.4's vocabulary.

- **Equal is legal**, which is what keeps the two-pass harvest/inject idiom
  working (corpus 03: a receiver's value read in one intervened model, injected
  at that same address in another).
- **Block-order aware for free**, because the comparison reuses the order group
  elision already depends on: a head's contribution feeding the same layer's MLP
  is upstream, and the reverse is refused. `plan._depth` is promoted to
  `plan.site_depth` rather than duplicated — one home for the order, two readers.
- Params and literal-scalar operands have no address, so they are unconstrained.

## Blast radius

It is a new refusal, so this is the claim that matters: **all 14 corpus
documents are already upstream-or-equal** — 20 operand edges, 4 upstream and 16
equal, 0 downstream. Pinned by a test that reads the authored JSON directly, so
it stays a statement about the corpus rather than about one point of an
expansion.

Full suite: 2815 passed, 27 skipped.

## The escape, for a value that really is a constant

The refusal says it: if the operand is meant as an externally supplied constant
rather than a routed activation, harvest it into a bundle in its own run and load
it as a `params` entry (§2.6) — where being a constant is the declared thing,
instead of being smuggled in as an activation read from the wrong end of the
network.

## Verification

```
$ causalab validate nonphysical_im.json     # sender L20 -> receiver L11
refused: [V20] at writes.e7_on.do write 'e7_on' takes its operand from read
'r7_cf' at site 's7' (attention_result layer 20), which is strictly deeper in
the forward pass than the address it lands on, site 'recv' (block_mid layer 11).
...
$ causalab validate circuit_im.json         # the same document, senders below
OK: 1 point, digest 0f1875de7298420f…
```

Nine tests cover: downstream refused, upstream legal, equal legal, block-order
awareness within one layer (both directions), `lm_head` sorting after every
block, params/literal operands unconstrained, the `alpha` slot constrained too,
and the corpus-wide geometry.

### DIFF
--- a/causalab/protocol/errors.py
+++ b/causalab/protocol/errors.py
@@ -35,7 +35,7 @@ __all__ = [
 #: How many rules the §5 load-error checklist has. Named so the range guard
 #: below and the spec move together: adding a rule means editing §5 and this
 #: number, and nothing else.
-CHECKLIST_RULES: int = 19
+CHECKLIST_RULES: int = 20
 
 
 class ProtocolError(ValueError):
--- a/causalab/protocol/plan.py
+++ b/causalab/protocol/plan.py
@@ -38,6 +38,7 @@ __all__ = [
     "interned_groups",
     "plan_point",
     "closure_digest",
+    "site_depth",
 ]
 
 #: Intra-block execution order of the component vocabulary — engine-free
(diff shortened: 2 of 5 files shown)


can-goodfire:

Moved to goodfire-ai/causalab-internal#562, rebased onto internal `main` (@ 99233b18, well ahead of this mirror) and re-verified there rather than assumed to port: the full internal suite passes, and every empirical claim in the description was re-measured against that tree.

Apologies for the noise.

Issue #79, opened 2026-09-01
A `block_mid` write is half-applied: it reaches the MLP but not the residual skip


can-goodfire:
## What

A write at `block_mid` lands at its own tap — a later read of `block_mid` in the
same intervened model returns the written value — but it does **not** reach the
residual skip that forms `block_output`. §2.4 states the identity

> `block_output = block_mid + mlp_output`

and under a write at `block_mid` that identity silently fails. The write moves
what `post_attention_layernorm` (hence the MLP) consumes, and leaves the skip on
the pre-write value.

## Measured (tiny-random-LlamaForCausalLM, layer 1, pos -1, fp32)

Writing `{"swap": v_cf}` at `block_mid`, all reads in the same model:

| check | result |
|---|---|
| `block_mid` read back == the written value | **True** (the write is not lost) |
| `block_output - mlp_output` == the **written** `block_mid` | **False**, max abs diff **8.5e-2** |
| `block_output - mlp_output` == the **unwritten** base `block_mid` | **True**, max abs diff 9.3e-10 |

## Why it went unnoticed

The sibling identity `block_output == block_input + attention_output +
mlp_output` is **blind** to this: a `block_mid` write touches neither
`block_input` nor `attention_output`, so it holds whether or not the skip was
updated. Only the two-term form catches it.

## Cause

HF's decoder layer captures the skip from the same tensor *before* the norm runs:

```python
residual = hidden_states                                  # skip captured here
hidden_states = self.post_attention_layernorm(hidden_states)
hidden_states = self.mlp(hidden_states)
hidden_states = residual + hidden_states                  # pre-write value
```

`block_mid` is the norm's *input*, so a pre-hook returning a new tensor cannot
reach the local `residual`.

## Possible fix

Lower a `block_mid` write into the equivalent write at `attention_output`, which
is a clean write point and already propagates correctly:

- additive: `block_mid += d`  ⟺  `attention_output += d`
- absolute: `block_mid ← V`   ⟺  `attention_output ← V - block_input`

Alternatively make `block_mid` read-only, as `attention_result`,
`router_logits` and `input_ids` already are — refusing beats silently
half-applying.

## Second, smaller defect

`block_mid` on GPT-2 raises `AttributeError: 'GPT2Block' object has no attribute
'post_attention_layernorm'` (it is `ln_2` there) — a crash rather than a clean
refusal, and the block has the same structure, so it looks resolvable rather
than genuinely absent.

## Pinned by

[...]

PR #81, opened 2026-09-01
Corpus 15: circuit evaluation over a per-edge polarity map, closed against the write it reduces to

can-goodfire:
## Why

`causalab-internal#534` argued against the *old* tree that its path-patching
method could not express a paper's circuit evaluation: it was
single-sender/all-others-frozen, and N single-edge passes are not the subgraph
pass, because the metric is nonlinear and the paths interact. That diagnosis was
right about the old API, and it is dissolved by this one — §2.8 allows unbounded
additive writes at one address, so an arbitrary per-edge polarity map is one
intervened model and one joint forward.

Nothing in the corpus said so. Corpus 15 does.

## The idiom

An **edge** is one head's contribution to a residual receiver:

| routing | how it is written |
|---|---|
| to counterfactual | `+1` on the counterfactual read of `attention_result`, `-1` on the base read |
| to base | absent |

Three name-mover edges are the circuit, one off-circuit head is the control, and
`explain` derives **four forwards** for the whole thing — two cache passes plus
one joint forward per edge set, not one per edge. No existing digest moved. The
document is also *run* end to end against the two-layer fixture, asserting that
routing moves the metric and that the two edge sets are distinguishable from
each other and from the clean run — a run that quietly wrote nothing would
produce three identical columns.

## The closure test is what makes it a claim

`attention_output == Σ_h attention_result + o_proj_bias`, so routing *every* head
must be indistinguishable from swapping the whole mixer output. It is, on both
families.

- **patch-everything** closes against the mixer swap.
- **patch-nothing** closes against the clean run, through a live `+1/-1` pair
  that sums to zero rather than around the machinery.
- **one edge ≠ the whole mixer** keeps the other two honest. Without it the
  closures could pass for a reason unrelated to per-edge routing.
- **a nonzero o-projection bias** leaks no term into the closure. Both fixtures
  dodge the bias otherwise — llama's projection has none and tiny-random gpt2's
  is all zeros — so this one sets a real one and restores it.

On the bias, stated precisely (an earlier commit message here overclaimed it,
and the third commit corrects it): the bias is the same constant on both inputs,
so it cancels in the difference **however the per-head derivation attributes
it**. The closure is therefore *insensitive* to attribution, not a test of it —
attribution is pinned by
`test_sites_round2_result.py::test_the_bias_belongs_to_no_head`, which already
did that job.

Unlike the analytic cache-arithmetic engine that PR proposed, nothing here
assumes the trunk is additive downstream of the receiver — everything past it is
a real forward. So there are no construction guards to maintain, and MoE /
DeltaNet layers are not special cases.

## A bug fell out

Writing the closure found one: **a `block_mid` write is half-applied.** It lands
at its own tap (a later read returns the written value) but not at the residual
skip, so §2.4's `block_output = block_mid + mlp_output` silently fails under it.
Measured on tiny-random llama, layer 1: `block_output - mlp_output` matches the
*unwritten* `block_mid` to 9.3e-10 and misses the written one by 8.5e-2.

The sibling identity `block_output == block_input + attention_output +
mlp_output` is **blind** to this — a `block_mid` write touches neither of the
first two terms — which is presumably why it survived.

The receiver here is `block_input` (corpus 03's own receiver shape) because of
it. The defect is pinned by a strict xfail and filed as #79, along with a second,
smaller one: `block_mid` on GPT-2 raises `AttributeError` rather than refusing
cleanly, since the module is `ln_2` there.

Full suite: 2819 passed, 27 skipped, 1 xfailed.

### DIFF
--- a/tests/neural/engines/pytorch_hooks/test_edge_delta_closure.py
+++ b/tests/neural/engines/pytorch_hooks/test_edge_delta_closure.py
@@ -14,10 +14,16 @@ trunk being linear all the way to the logits:
 
 so routing *every* head of layer L into the mixer output itself must land it
 on ``Σ_h result_cf + bias`` — which is ``attention_output`` on the
-counterfactual, exactly what swapping the whole mixer output produces. The
-bias belongs to no head and enters once, from the base value, so it cancels;
-the ``bundle`` fixture covers gpt2, whose o-projection *has* a bias, as well
-as llama, whose does not.
+counterfactual, exactly what swapping the whole mixer output produces.
+
+The o-projection bias belongs to no head, and it drops out for a reason worth
+stating precisely: it is the *same constant* in the base and counterfactual
+values, so it cancels in the difference **however the derivation attributes
+it** — the closure is insensitive to bias attribution rather than a test of it.
+What pins the attribution itself is
+``test_sites_round2_result.py::test_the_bias_belongs_to_no_head``. What the
+nonzero-bias test below adds is that a real bias, rather than the fixtures'
+absent (llama) or all-zero (gpt2) one, leaks no term into the closure.
 
 The receiver is ``attention_output`` and not ``block_mid``: a ``block_mid``
 write is currently only half-applied (it reaches the MLP but not the residual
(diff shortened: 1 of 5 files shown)


can-goodfire:

Moved to goodfire-ai/causalab-internal#564, rebased onto internal `main` (@ 99233b18, well ahead of this mirror) and re-verified there rather than assumed to port: the full internal suite passes, and every empirical claim in the description was re-measured against that tree.

Apologies for the noise.

PR #82, opened 2026-09-01
docs: the per-head q/k/v receiver gaps are closed, not open

can-goodfire:
## Why

Three statements in `docs/test_migration.md` still describe the component
vocabulary as it was before round 2's attention interior landed. All three are
now wrong, and one of them says a capability is *inexpressible* when it ships and
runs — the kind of stale note that makes someone rebuild what already exists.

## What changed

| where | said | actually |
|---|---|---|
| retired-test table | per-head value and query receivers "have no §2.4 component" | `attention_value_states` is v_proj's output in KV-head space; `attention_query_pre_rope` is the pre-attention q_h = W_Q^h·x. Neither is `attention_premix`, the o_proj *input* in query space — the confusion the row was recording |
| known-gaps list | path patching onto query receivers "is inexpressible in v1" | expressible, and it runs |
| parity-goldens entry | 2 gqa `head_value` pins skipped because "the vocabulary has no v_proj-output component" | that reason is gone; the skips are not |

The parity-goldens entry stays a gap — I only corrected its reason. What the
pins now need is a re-audit of whether the retired oracle's head indexing matches
the KV-head space the component presents under GQA, which is unmeasured. Better
to record the real open question than a resolved one.

## Verified, not reasoned

All three components were driven as per-head receivers through corpus 03's
two-pass shape on the reference engine (tiny-random llama, layer 1), each
producing finite logits:

```
attention_query_pre_rope   head=1 -> RAN, logits (1, 1, 32000), finite=True
attention_value_states     head=0 -> RAN, logits (1, 1, 32000), finite=True
attention_key_pre_rope     head=0 -> RAN, logits (1, 1, 32000), finite=True
```

The GQA head bound is enforced separately, and refuses in the component's own
space rather than the model's: `head: 20` on `attention_key_pre_rope` gives
`[V4] … head 20 out of range (8 heads in 'attention_key_pre_rope''s head space)`
on a 32-query-head model.

## Context

Noticed while checking whether `causalab-internal#534`'s `head_key_input`
receiver had anything left to port. It does not — this tree has the whole q/k/v
surface, pre- and post-RoPE, with head bounds derived from component shape rather
than hand-written.

Docs only; no code changes.

### DIFF
--- a/docs/test_migration.md
+++ b/docs/test_migration.md
@@ -82,7 +82,7 @@ file* paths are relative to `tests/`. Parametrized data dirs (`configs/`,
 | `interchange/test_layer_scan.py` | replaced interface | `protocol/test_sweep.py`, `protocol/test_corpus.py` (corpus 07) | the layer scan is a sweep axis; deterministic expansion + the locate-grid shape pinned |
 | `interchange/test_single_pair.py` | replaced interface | `neural/pytorch_hooks/test_positions_frame.py::test_out_of_bounds_refuses` | the #176 contract carried verbatim in spirit: a stale position refuses, never addresses the wrong token |
 | `path_patching/conftest.py` | retired | retired | Plan-era fixtures (single-step pipelines, pyvene hook-cleanup workarounds) |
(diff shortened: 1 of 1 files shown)


can-goodfire:

Moved to goodfire-ai/causalab-internal#565, rebased onto internal `main` (@ 99233b18, well ahead of this mirror) and re-verified there rather than assumed to port: the full internal suite passes, and every empirical claim in the description was re-measured against that tree.

Apologies for the noise.

Issue #38, opened 2026-08-24
Dangling references to the deleted `causalab/runner/` package (one in user-facing error text)


can-goodfire:
## Context

Found while driving a full analysis pipeline against the protocol refactor
(#20) on a real task. These are not design gaps — they are references left
pointing at `causalab/runner/`, a package the refactor deletes. One of them is
**runtime error text shown to a user**, one is a live feature whose only
documentation was deleted with it, and one is a guard that silently stopped
guarding.

Verified against `can/protocol-refactor` @ `40dafcd` (the tip with #21, #27,
#28, #29, #30 merged). `causalab/runner/` does not exist there.

### 1. A user-facing error points at a deleted file — `causalab/tasks/loader.py:203`

```python
raise ModuleNotFoundError(
    f"No task package {task_name!r} found. Tried: {', '.join(candidates)}. "
    f"For session-local tasks, see causalab/runner/README.md "
    f"'Session-local code injection'."
)
```

Anyone who mistypes a task name is told to read a file that is not in the
repository. This is the one with a real cost.

### 2. `CAUSALAB_SESSION_CODE` survived; its documentation did not

`causalab/tasks/loader.py:158` still reads the variable, so session-local task
injection **is a live feature** — but its only description was
`causalab/runner/README.md` § "Session-local code injection". Three docstrings
now forward to it (`:152`, `:182`, `:189`, the last also citing
`causalab.runner.run_exp` for its precedence rule).

Either the feature keeps its documentation somewhere that exists, or — if
session-local injection is not meant to survive into the protocol era — it
should be removed deliberately rather than left working and undescribed. That
is a decision this issue asks for, not one it makes.

### 3. An orphaned renderer — `causalab/io/plots/task_setup.py:11`

The docstring says "the thin runner CLI `causalab.runner.task_setup_figure`
loads the tasks, samples the examples, and calls `write_task_setup_html`".
That CLI is deleted. `write_task_setup_html` still exists and is still tested
(`tests/io/plots/test_task_setup.py`), but **nothing in `causalab/` calls it** —
it is a renderer with no producer. The artifact viewer still expects its output
(`tests/io/test_viewer_spec_merge.py:195` lists `plan/figures/task_setup.html`
as a viewer candidate), so the figure is still wanted; only the thing that
generated it is gone.

### 4. A layering guard that can no longer fail — `tests/test_architecture_layering.py:28`

```python
[...]

PR #83, opened 2026-09-16
docs: repair references to retired runner package

nehal-a2z:
Mistyped task names currently send users to the deleted runner README. Point that diagnostic and the loader docstrings to a new session-local task section in the surviving task README, documenting the existing environment gate, caller-owned import path, and shipped-first precedence.

Also correct the remaining I/O and task-setup renderer descriptions: the layering guard already checks the current workflow packages, and the renderer currently has no shipped producer. Add a guard against retired runner references under `causalab/` and regression coverage for the diagnostic's documentation target and session gate. Runtime task resolution, rendering, and the existing layering policy are unchanged.

Addresses #38. This documents the surviving task-loading feature and the renderer's present status; it does not restore the retired orchestration stack or add a figure producer.

Validation:

- New diagnostic/reference tests reproduced the failures before the fix.
- 16 focused loader/session/architecture tests passed with `--noconftest` because local PyTorch is unavailable. This bypasses the root fixtures/hooks and tier enforcement; a separate collection audit verified one valid tier on each selected test (11 property, 5 unit), and these tests use no root fixtures or golden teardown.
- Real dataset-builder smoke produced 64 rows from a session-local singleton task.
- Ruff 0.14.5 check and format check passed on changed Python files; `git diff --check` passed.
- The unchanged renderer suite could not collect locally because the plots package imports PyTorch. The full suite has not run locally.
- [CI run 35096174486](https://github.com/goodfire-ai/causalab/actions/runs/35096174486) stopped before dependency installation/tests: `actions/checkout@v5` refuses fork code in the existing privileged `pull_request_target` workflow. Fork lint is skipped. No workflow permissions or checkout safeguards are changed by this patch.


can-goodfire:
Thanks for your fixes @nehal-a2z , we've been working on a newer version in parallel. Would you be able to check whether the issues you've identified persist in the new version we've just merged into main?

